Issue #193, opened 2015-01-29
"Component: Sl-Menu: Menu properly hides all but root list" test passes in browser but not CLI


notmessenger:
actual: >
            11
        expected: >
            3
        Log: >

notmessenger:
Resolved via https://github.com/softlayer/sl-ember-components/pull/197

Issue #196, opened 2015-01-29
"Component: Sl-Menu: Menu selection fires proper selection event" test passes in browser but not CLI


notmessenger:
actual: >
            2
        expected: >
            3
        Log: >

notmessenger:
Resolved via https://github.com/softlayer/sl-ember-components/pull/197

Issue #195, opened 2015-01-29
"Component: Sl-Menu: Menu closes child on mouse exit" test passes in browser but not CLI


notmessenger:
actual: >
            11
        expected: >
            3
        Log: >

notmessenger:
Resolved via https://github.com/softlayer/sl-ember-components/pull/197

PR #197, opened 2015-01-29
Change assets names from "sl-ember-components" to "dummy"

notmessenger:


notmessenger:
Fixes #193, #194, #195, #196

PR #335, opened 2015-05-18
Close #329 - Add ability to set class attributes on sl-menu components

### LINKED ISSUE #329: Support ability to apply class to a root-level menu option based on provided parameter
This will allow for an application to pass this property into the menu based on current route or some other criteria and have a class applied to the appropriate menu for styling purposes.

jonathandavidson:
Custom classes can be added to an sl-menu's element by including an optional classNames property to the menu object. The value should be an array of strings, each of which will be pushed to the corresponding sl-menu's classNames attribute, which is bound to that menu's element's class attribute.

### DIFF
--- a/addon/components/sl-menu.js
+++ b/addon/components/sl-menu.js
@@ -338,6 +338,13 @@ export default Ember.Component.extend({
     children: null,
 
     /**
+     * Additional class names to be added to the classNames attribute
+     *
+     * @type {String[]}
+     */
+    extraClassNames: [],
+
+    /**
      * @property {boolean} isRoot
      * @default  true
      */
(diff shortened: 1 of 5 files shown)


addon/components/sl-menu.js
```diff
@@ -396,6 +403,22 @@ export default Ember.Component.extend({
     }.on( 'init' ),
 
     /**
+     * Initialize initClassNames array
+     *
+     * @function initClassNames
+     * @observes "init" event
```
notmessenger: should be: @observes init

addon/components/sl-menu.js
```diff
@@ -396,6 +403,22 @@ export default Ember.Component.extend({
...
+     * @returns  {undefined}
```
notmessenger: Only need one space between the tag and type

addon/components/sl-menu.js
```diff
@@ -396,6 +403,22 @@ export default Ember.Component.extend({
...
+     */
+    initClassNames: function() {
+        var classNames = this.get( 'classNames' ),
```
notmessenger: align the equal sign with the one on the line below it

addon/components/sl-menu.js
```diff
@@ -396,6 +403,22 @@ export default Ember.Component.extend({
...
+            extraClassNames = this.get( 'extraClassNames' );
+
+        if ( extraClassNames && extraClassNames.length > 0 ) {
```
notmessenger: The check for extraClassNames should not be necessary since it has already been initialized as an array.  Regardless, this entire line should be replaced with an Ember.isNone/isEmpty or other call as appropriate.

addon/components/sl-menu.js
jonathandavidson: Per our discussion we are standardizing on not aligning var assignments equal signs.

PR #337, opened 2015-05-20
Close #336 - sl-menu custom classes are set on init, but need to be bound to allow changing after initialization

### LINKED ISSUE #336: sl-menu custom classes are set on init, but need to be bound to allow changing after initialization

jonathandavidson:


addon/components/sl-menu.js
```diff
@@ -20,6 +20,9 @@ export default Ember.Component.extend({
      */
     tagName: 'div',
 
+    /** @type {Array} */
```
notmessenger: type is String[]

tests/unit/components/sl-menu-test.js
```diff
@@ -7,7 +7,7 @@ var clickCounter = 0,
     modelStub = {
         label: null,
         pages: [
-            { label: 'Top Level A', classNames: [ 'class1', 'class2' ], pages: [
+            { label: 'Top Level A', className: 'class1 class2', pages: [
```
notmessenger: Prefer that these be specified as an array rather than a text string.  It will be easier to dynamically manipulate an array of options passed to this component than a text string.

tests/dummy/app/routes/demos/sl-menu.js
```diff
@@ -5,7 +5,7 @@ export default Ember.Route.extend({
 
     model: function() {
         return  Ember.Object.create({ label: null, pages: [
-            Ember.Object.create({ label: 'Colors', classNames: [ 'current' ], pages: [
+            Ember.Object.create({ label: 'Colors', className: 'current', pages: [
```
notmessenger: Prefer that these be specified as an array rather than a text string. It will be easier to dynamically manipulate an array of options passed to this component than a text string.

tests/dummy/app/templates/demos/sl-menu.hbs
```diff
@@ -60,7 +60,7 @@
         <ul>
             <li><strong>pages</strong> - An array of nodes that will be embedded under the <code>label</code> attribute as a sub-menu.</li>
             <li><strong>label</strong> - The label of the menu itself. This should be <em>null</em> for top level menus and will provide the name of the sub-menu for sub-menu lists.</li>
-            <li><strong>classNames</strong> - An array of class names to be added to the element's <code>class</code> attribute.</li>
+            <li><strong>className</strong> - An string of classes to be added to the element's <code>class</code> attribute.</li>
```
notmessenger: Prefer that these be specified as an array rather than a text string. It will be easier to dynamically manipulate an array of options passed to this component than a text string.

addon/components/sl-menu.js
```diff
@@ -338,11 +341,11 @@ export default Ember.Component.extend({
...
 
     /**
-     * Additional class names to be added to the classNames attribute
+     * Additional class string to be added to the element's class attribute
      *
-     * @type {String[]}
+     * @type {String}
      */
-    extraClassNames: [],
+    className: '',
```
notmessenger: Change this name to something not so similar to the "classNames" property that already exists on components.

addon/components/sl-menu.js
```diff
@@ -338,13 +341,28 @@ export default Ember.Component.extend({
...
+     * Array of classes to be added to the element's class attribute
      *
      * @type {String[]}
      */
     extraClassNames: [],
 
     /**
+     * Additional class string to be added to the element's class attribute
+     *
+     * @type {String}
```
notmessenger: doc block tags should be:

@function
@returns {String|Array}

addon/components/sl-menu.js
```diff
@@ -338,13 +341,28 @@ export default Ember.Component.extend({
...
+     */
+    extraClassNamesString: function() {
+        var extraClassNames = this.get( 'extraClassNames' );
+
+        if ( !Ember.isNone( extraClassNames ) ) {
+            return extraClassNames.join( ' ' );
+        } else {
+            return [];
```
notmessenger: Don't think it really matters but do we want this to return an array or just an empty string.  Better yet maybe, only return a value if there were names to join?  Not sure what the binding to an "undefined" value does if approach it that way.

jonathandavidson: Oh that's a mistake. It should return an empty string if there are no extraClassNames.

notmessenger: Cool.

The preferred pattern at that point then is to define a return value at the top and return it at the end of the function.  The if statement can modify it as appropriate.  Falls into the "only one return statement per method" recommendation.

addon/components/sl-menu.js
```diff
@@ -338,13 +341,29 @@ export default Ember.Component.extend({
...
      * @type {String[]}
      */
     extraClassNames: [],
 
     /**
+     * Additional class string to be added to the element's class attribute
+     *
+     * @type {String}
+     */
+    extraClassNamesString: function() {
```
notmessenger: This should be in the "Methods" section labeled as a @function with @returns {String}

PR #343, opened 2015-06-03
spinnerFix - created a variables.less, updated path to spinner-dark/spinner-light to variable definition

juwara0:
Created a variables.less file containing variables with the path to the spinner-dark and spinner-light png files. Updated the places where spinner-dark/spinner-light png's were used and imported new variables.less file into the sl-ember-components.less file.


azizpunjani:
Maybe it would also make sense to move out the base URL path into its own variable? `@baseUrlPath: 'sl-ember-components/assets/'`. I noticed a few places in the repo that we use this path, it would be nice to keep things DRY. I can open a separate PR for this if needed.

notmessenger:
Neither of these approaches are desired.

The way the pathing is currently setup is according to how Ember CLI Addons are configured and everything works as expected.  The reason this even came up is because SoftLayer is serving their Ember application from a location on disk different than the web root of their web server configuration and it is messing up the pathing.  But this is a problem to be solved in their application and not by making non-standard concessions in this addon.

(Yes, I know that I am representing SoftLayer when I am writing this comment and I'm also referring to SoftLayer as a separate entity but I did so to help illustrate and separate the pieces in play)

Issue #6, opened 2014-11-21
Pressing the "-" key more times than expect user to causes errors


notmessenger:
Uncaught TypeError: undefined is not a function

sl-components/components/sl-menu.js:79 - child.drillDown() call in drillDown()

notmessenger:
Resolved by https://github.com/softlayer/sl-ember-components/pull/499

PR #355, opened 2015-06-15
Fixed bug when double -- is pressed

azizpunjani:


azizpunjani:
Addresses https://github.com/softlayer/sl-ember-components/issues/6

azizpunjani:
Also updated doc block comment to be more accurate.

notmessenger:
When I created this issue I believed that the occurrence of this error was what was preventing the 3rd level of menus from being selected (e.g. Cities > Texas > Dallas).  While this PR keeps the error from occurring it did not resolve this menu selection problem.  I'm torn on merging this PR in and then creating a separate issue for the scenario I just described as it is quite likely that these two are related to one another.  Do you mind doing a little research into this new development and then we can determine whether it is related or should be addressed as a separate issue if will need its own development effort?

azizpunjani:
Will investigate further to see why the selection is not working. I did test the scenario you mentioned but that was in control and it did work.

azizpunjani:
@notmessenger Is this issue now obsolete because of https://github.com/softlayer/sl-ember-components/issues/359 ?

notmessenger:
@azizpunjani Was wondering the same thing.  @joshforisha can you please weigh in on this?

joshforisha:
I hadn't intended to change too much of the sl-menu component for that issue, but it turns out it needs quite a bit of cleanup. I'm fairly positive at this point it will obsolete this PR and issue, though it depends on the speed of the menu refactor.

Issue #354, opened 2015-06-15
addon/templates/components/sl-progress-bar.hbs is binding unescaped style attributes


notmessenger:
WARNING: Binding style attributes may introduce cross-site scripting vulnerabilities; please ensure that values being bound are properly escaped. For more information, including how to disable this warning, see http://emberjs.com/deprecations/v1.x/#toc_binding-style-attributes.

http://emberjs.com/deprecations/v1.x/#toc_binding-style-attributes

  style={{styleString}}

notmessenger:
Resolved by https://github.com/softlayer/sl-ember-components/pull/360

PR #360, opened 2015-06-17
Make size a SafeString when size is not a number

azizpunjani:
Fix for
WARNING: Binding style attributes may introduce cross-site scripting vulnerabilities; please ensure that values being bound are properly escaped. For more information, including how to disable this warning, see http://emberjs.com/deprecations/v1.x/#toc_binding-style-attributes.


azizpunjani:
Addresses https://github.com/softlayer/sl-ember-components/issues/354

Issue #26, opened 2014-11-25
Missing tests for default values


notmessenger:
New default values were added to the component but the tests were not updated.

Issue #245, opened 2015-02-04
Add tests for sl-tooltip-enabled mixin


notmessenger:

joshforisha:
Working on this.

notmessenger:
https://github.com/softlayer/sl-ember-components/issues/26 seems related

PR #377, opened 2015-06-25
Closes softlayer/sl-ember-components#26,#245

juwara0:
Added tests for sl-tooltip and asserts for verifying default types. This work was being tracked as issues: 26 and 245.


notmessenger:
Based on in-person conversation with @juwara0 this PR is being closed pending a few modifications

PR #379, opened 2015-06-26
Closes softlayer/sl-ember-components#26,#245

juwara0:
Closes softlayer/sl-ember-components#26,#245

### DIFF
--- a/addon/components/sl-tooltip.js
+++ b/addon/components/sl-tooltip.js
@@ -15,7 +15,7 @@ export default Ember.Component.extend( TooltipEnabled, {
     // Attributes
 
     /** @type {String} */
-    tagName: 'span'
+    tagName: 'span',
 
     // -------------------------------------------------------------------------
     // Actions
@@ -28,6 +28,26 @@ export default Ember.Component.extend( TooltipEnabled, {
 
     // -------------------------------------------------------------------------
     // Observers
+
+    /**
+     * Check passed parameters on initialization
+     *
+     * @function
+     * @returns {undefined}
+     */
+    initialize: Ember.on (
+        'init',
+        function() {
+            Ember.assert(
+                'method/enablePopover() expects the parameter "title" and for it to be a string',
+                'string' === Ember.typeOf ( this.get( 'title' ) )
+            );
+            Ember.assert(
+                'method/enablePopover() expects the parameter "popover" and for it to be a string',
+                'string' === Ember.typeOf ( this.get( 'popover' ) ) || 'undefined' === Ember.typeOf ( this.get( 'popover' ) )
+            );
+        }
+    )
 
     // -------------------------------------------------------------------------
     // Methods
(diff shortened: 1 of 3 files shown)


addon/components/sl-tooltip.js
```diff
@@ -29,6 +29,26 @@ export default Ember.Component.extend( TooltipEnabled, {
     // -------------------------------------------------------------------------
     // Observers
 
+    /**
+     * Check passed parameters on initialization
+     *
+     * @function
```
notmessenger: Add @throws

addon/components/sl-tooltip.js
```diff
@@ -29,6 +29,26 @@ export default Ember.Component.extend( TooltipEnabled, {
...
+    initialize: Ember.on (
+        'init',
+        function() {
+            Ember.assert(
+                'method/enablePopover() expects the parameter "title" and for it to be a string',
+                'string' === Ember.typeOf ( this.get( 'title' ) )
+            );
+            Ember.assert(
+                'method/enablePopover() expects the parameter "popover" and for it to be a string',
+                'string' === Ember.typeOf ( this.get( 'popover' ) ) || 'undefined' === Ember.typeOf ( this.get( 'popover' ) )
```
notmessenger: The second condition will never be reached because the first one will catch the string of "undefined"

addon/components/sl-tooltip.js
```diff
@@ -29,6 +29,26 @@ export default Ember.Component.extend( TooltipEnabled, {
...
+     *
+     * @function
+     * @returns {undefined}
+     */
+    initialize: Ember.on (
+        'init',
+        function() {
+            Ember.assert(
+                'method/enablePopover() expects the parameter "title" and for it to be a string',
+                'string' === Ember.typeOf ( this.get( 'title' ) )
```
notmessenger: This will not handle `undefined` values

tests/unit/components/sl-tooltip-test.js
```diff
@@ -8,7 +8,77 @@ moduleForComponent( 'sl-tooltip', 'Unit | Component | sl tooltip', {
 
 test( 'Expected Mixin is present', function( assert ) {
     assert.ok(
-        TooltipEnabledMixin.detect( this.subject() ),
+        TooltipEnabledMixin.detect( this.subject( { title: 'Tooltip Text' } ) ),
         'Expected Mixin is present'
     );
 });
+
+test( 'enabledTooltip() - Verify classes are set correctly: data-toggle and data-original-title', function( assert ) {
```
notmessenger: The name of the test does not need to be so specific (the inclusion of the specific class names).  This is so that there are less things to be out of sync with the code if a class is added or removed in the future.

notmessenger:
Per our conversation about these tests not belonging on this component and instead on the mixing I am closing this PR

PR #382, opened 2015-06-30
Closes softlayer/sl-ember-components#26

juwara0:
Moved asserts to an init method, modified tests for sl-tooltip

### DIFF
--- a/tests/unit/components/sl-tooltip-test.js
+++ b/tests/unit/components/sl-tooltip-test.js
@@ -137,6 +137,24 @@ test( '"title" property needs to be a string', function( assert ) {
         'Property was an Object'
     );
 
+    // Undefined Property
+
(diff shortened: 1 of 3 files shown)


tests/unit/components/sl-tooltip-test.js
```diff
@@ -8,7 +8,294 @@ moduleForComponent( 'sl-tooltip', 'Unit | Component | sl tooltip', {
...
+        this.$().prop( 'tagName' ),
+        'SPAN'
+    );
+    assert.equal(
+        this.$().prop( 'class' ),
+        'ember-view'
+    );
+});
+
+test( '"title" property needs to be a string', function( assert ) {
```
notmessenger: an assertion for an undefined property should also be ran

PR #390, opened 2015-07-06
fixes bug where sl-select couldn't recognize Ember objects passed in 

JKGisMe:
If an array of Ember objects (ie array of records from the store) is passed in for the `content` property sl-select couldn't recognize them because `Ember.typeOf( item )` returns 'instance' for an Ember object. The conditionals only looked for type 'object'.

I didn't implement any tests or anything. Just offering what worked for me.

### DIFF
--- a/addon/components/sl-select.js
+++ b/addon/components/sl-select.js
@@ -129,7 +129,7 @@ export default Ember.Component.extend( InputBased, TooltipEnabled, {
                         return;
                     }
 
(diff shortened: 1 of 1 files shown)


notmessenger:
@JKGisMe Please submit either https://github.com/softlayer/sl-ember-components/blob/develop/CLA-INDIVIDUAL.md or https://github.com/softlayer/sl-ember-components/blob/develop/CLA-CORPORATE.md and let me know once you have so that I can merge this PR.  Thanks!

Issue #207, opened 2015-02-04
Add tests for sl-chart component


notmessenger:

erangeles:
grabbing this, @notmessenger when u see this can you assign me it. Thanks

notmessenger:
Resolved by https://github.com/softlayer/sl-ember-components/pull/406

PR #395, opened 2015-07-07
issue #207, Add tests for sl-chart component

erangeles:
initial set of tests added for the sl-chart component.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,7 +4,9 @@ There are many ways to contribute to the project, including fixing issues or imp
 
 # Questions
 
-This is the issue tracker for `sl-ember-components`. The community uses this site to collect and track bugs and discussions of new features. If you are having difficulties using `sl-ember-components` or have a question about usage please ask a question on Stack Overflow: http://stackoverflow.com/questions/ask?tags=sl-ember-components
+This is the issue tracker for `sl-ember-components`. The community uses this site to collect and track bugs and
+discussions of new features. If you are having difficulties using `sl-ember-components` or have a question about
+usage please ask a question on Stack Overflow: http://stackoverflow.com/questions/ask?tags=sl-ember-components
 
 # Issues
 
@@ -12,18 +14,20 @@ Think you've found a bug or have a new feature to suggest? Let us know!
 
 ## Reporting a Bug
 
-1. Make sure you have the latest version of the code, if possible, as we may have already fixed your bug.  This project uses the [git-flow branching model](https://github.com/nvie/gitflow) so the latest changes will be in the `develop` branch.
+1. Make sure you have the latest version of the code, if possible, as we may have already fixed your bug.  This
+project uses the [git-flow branching model](https://github.com/nvie/gitflow) so the latest changes will be in the `
+develop` branch.
 
 2. Search for similar issues. It's possible somebody has encountered this bug already.
 
-3. Provide a demo that specifically shows the problem. This demo should be fully operational with the exception of the
-bug you want to demonstrate. The more pared down, the better. Issues with demos are prioritized.
+3. Provide a demo that specifically shows the problem. This demo should be fully operational with the exception of
+the bug you want to demonstrate. The more pared down, the better. Issues with demos are prioritized.
 
-4. Your issue will be verified. The provided demo will be tested for correctness. The sl-ember-components team will work with
-you until your issue can be verified.
+4. Your issue will be verified. The provided demo will be tested for correctness. The sl-ember-components team will
+work with you until your issue can be verified.
 
-5. Keep up to date with feedback from the sl-ember-components team on your ticket. Your ticket may be closed if it becomes
-stale.
+5. Keep up to date with feedback from the sl-ember-components team on your ticket. Your ticket may be closed if it
+becomes stale.
 
 6. If possible, submit a Pull Request with a failing test. Better yet, take
 a stab at fixing the bug yourself if you can!
@@ -36,12 +40,12 @@ there is a bug and the faster we'll be able to take action.
 1. Search Issues for similar feature requests. It's possible somebody has already asked
 for this feature or provided a pull request that we're still discussing.
 
-2. Provide a clear and detailed explanation of the feature you want and why it's important to add. Keep in mind that we
-want features that will be useful to the majority of our users and not just a small subset. If you're just targeting a
-minority of users, consider writing your own addon library for `ember-cli` that extends this one.
+2. Provide a clear and detailed explanation of the feature you want and why it's important to add. Keep in mind that
+we want features that will be useful to the majority of our users and not just a small subset. If you're just
+targeting a minority of users, consider writing your own addon library for `ember-cli` that extends this one.
 
-3. If the feature is complex, consider writing some initial documentation for it. If we do end up accepting the feature
-it will need to be documented and this will also help us to understand it better ourselves.
+3. If the feature is complex, consider writing some initial documentation for it. If we do end up accepting the
+feature it will need to be documented and this will also help us to understand it better ourselves.
 
 4. Attempt a Pull Request. If you are willing to help the project out, you can submit a Pull Request. We always have
 more work to do than time to do it. If you can write some code then that will speed the process along.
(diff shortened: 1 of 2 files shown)


tests/unit/components/sl-chart-test.js
```diff
@@ -0,0 +1,121 @@
...
+
+moduleForComponent( 'sl-chart', 'Unit | Component | sl chart', {
+	unit: true
+});
+
+test( 'Default classNames are present', function( assert ) {
+    this.subject({ object, series });
+
+    assert.ok(
+        this.$().hasClass('panel'),
```
notmessenger: Add space after open parenthesis and before closing one.  See https://github.com/softlayer/ember-style-guide/blob/master/javascript.md#block-statements

tests/unit/components/sl-chart-test.js
```diff
@@ -0,0 +1,121 @@
...
+        'Default rendered component has class "panel"'
+    );
+
+    assert.ok(
+        this.$().hasClass('panel-default'),
```
notmessenger: Add space after open parenthesis and before closing one.  See https://github.com/softlayer/ember-style-guide/blob/master/javascript.md#block-statements

tests/unit/components/sl-chart-test.js
```diff
@@ -0,0 +1,121 @@
...
+        'Default rendered component has class "panel-default"'
+    );
+    
+    assert.ok(
+        this.$().hasClass('sl-chart'),
```
notmessenger: Add space after open parenthesis and before closing one.  See https://github.com/softlayer/ember-style-guide/blob/master/javascript.md#block-statements

tests/unit/components/sl-chart-test.js
```diff
@@ -0,0 +1,121 @@
...
+        'Default rendered component has class "sl-chart"'
+    );
+    
+    assert.ok(
+        this.$().hasClass('sl-panel'),
```
notmessenger: Add space after open parenthesis and before closing one.  See https://github.com/softlayer/ember-style-guide/blob/master/javascript.md#block-statements

notmessenger:
Lines 42, 47 and 75 have empty spaces on them.  See https://github.com/softlayer/ember-style-guide/blob/master/javascript.md#whitespace

tests/unit/components/sl-chart-test.js
```diff
@@ -0,0 +1,121 @@
...
+        component.set( 'isLoading', true );
+    });
+
+    assert.ok(
+        this.$().hasClass( 'sl-loading' ),
+        'Rendered component gains "sl-loading" class'
+    );
+});
+
+test('setupChart() is called once', function(assert) {
```
notmessenger: Don't need to check that `setupChart()` is called once, specifically, because it will fire on the `didInsertElement` event.  What we should check though is that when the component is rendered (on/via `didInsertElement`) that this logic is being exercised as expected:

```
chartDiv.highcharts( options );
this.set( 'chart', chartDiv.highcharts() );
this.updateData();
```

tests/unit/components/sl-chart-test.js
```diff
@@ -0,0 +1,121 @@
+import Ember from 'ember';
+import { moduleForComponent, test } from 'ember-qunit';
+
+var object = {
```
notmessenger: Replace `var` with `let`.  See https://github.com/softlayer/ember-style-guide/blob/master/javascript.md#variables

tests/unit/components/sl-chart-test.js
```diff
@@ -0,0 +1,121 @@
...
+
+        yAxis: {
+            title: {
+                text: 'Fruit Eaten'
+            }
+        }
+    }
+};
+
+var series = Ember.A([
```
notmessenger: Replace `var` with `let`.  See https://github.com/softlayer/ember-style-guide/blob/master/javascript.md#variables

tests/unit/components/sl-chart-test.js
```diff
@@ -0,0 +1,121 @@
...
+    );
+    
+    assert.ok(
+        this.$().hasClass( 'sl-panel' ),
+        'Default rendered component has class "sl-panel"'
+    );
+});
+
+test( 'Loading state adds loading class', function( assert ) {
+    var component = this.subject({ object, series });
```
notmessenger: Replace `var` with `let`.  See https://github.com/softlayer/ember-style-guide/blob/master/javascript.md#variables

tests/unit/components/sl-chart-test.js
notmessenger: Not a fan of the use of `object` as it is not descriptive enough and is similar to a JavaScript concept.  See https://github.com/softlayer/ember-style-guide/blob/master/javascript.md#naming-conventions

`series` was used for the series data so `options` seems acceptable, or `testOptions`.  If select the latter or a differing pattern, then also align `series` to be similar.

tests/unit/components/sl-chart-test.js
```diff
@@ -0,0 +1,121 @@
...
+            options: object,
+            series: series
+        });
+    }  catch( error ) {
+        assertionThrown = true;
+    }
+
+    assert.ok(
+        !assertionThrown,
+        'proptery was an Object'
```
notmessenger: Spelling

tests/unit/components/sl-chart-test.js
```diff
@@ -0,0 +1,121 @@
...
+    this.render();
+
+    assert.equal(
+        spy.calledOnce,
+        true,
+        'setupChart() is called once'
+    );
+});
+
+test( '"Options" property needs to be an object', function( assert ) {
```
notmessenger: This test only confirms that passing an Object to `options` does not cause any errors.  What happens when you pass a value that is (a) `empty`, `Number`, `Boolean`, `Array`, `Function`, `undefined`, `String`?  Seems we need to add some assertion logic to `setupChart()` to enforce the acceptance of only Object types and expand this test.

tests/unit/components/sl-chart-test.js
notmessenger: Do not use `Ember.A()`, only `[]` in this situation.  The reason for this is because the code under test only needs an array - not an Ember Array specifically.  Aim to keep the contrived test scenarios as basic and pure as possible so that we limit the amount of things they can be affected by and when a test fails it should be very obvious as to why it did and not because we set the test up incorrectly by introducing unneeded "overhead".

tests/unit/components/sl-chart-test.js
erangeles: @notmessenger should I create an issue to add the assertion logic on huboard to track that change?

tests/unit/components/sl-chart-test.js
```diff
@@ -0,0 +1,121 @@
...
+        assertionThrown = true;
+    }
+
+    assert.ok(
+        !assertionThrown,
+        'proptery was an Object'
+    );
+});
+
+test( '"Series" property needs to be an array', function( assert ) {
```
notmessenger: This test only confirms that passing an Array to `series` does not cause any errors. What happens when you pass a value that is (a) `empty`, `Number`, `Boolean`, `Object`, `Function`, `undefined`, `String`, or 'Symbol`? Seems we need to add some assertion logic to the `init` event to enforce the acceptance of only Array types and expand this test.

notmessenger:
`style()` is not being tested and needs to be

notmessenger:
There are not tests regarding the `title` property and there needs to be

notmessenger:
This property can be removed from the component - https://github.com/softlayer/sl-ember-components/blob/develop/addon/components/sl-chart.js#L96

This is not in the test file but since the tests are being written to ensure the expected functionality of the component is solid, and in doing so this property was determined to be unneeded, go ahead and remove it as part of the same PR.

tests/unit/components/sl-chart-test.js
notmessenger: You can go ahead and do it as part of this PR.  Since the tests are being written to ensure the expected functionality of the component is solid any necessary changes to the component can be made now that are found to part of the fundamental architecture, for if the tests had been written when the component was originally written it would have been changed at that point without making it into the code.

erangeles:
@notmessenger this is ready to be reviewed and merged in.

Issue #240, opened 2015-02-04
Add tests for sl-input-based mixin


notmessenger:

alxyuu:
@notmessenger  picking this one up

notmessenger:
Resolved by https://github.com/softlayer/sl-ember-components/pull/405

PR #405, opened 2015-07-09
#240 add tests for sl-input-based mixin

alxyuu:


tests/unit/mixins/sl-input-based-test.js
```diff
@@ -0,0 +1,200 @@
...
+module( 'Unit | Mixin | sl input based' );
+
+test( 'Successfully mixed', function( assert ) {
+    let testObject = Ember.Object.extend( mixinUnderTest );
+    let subject = testObject.create();
+
+    assert.ok( subject );
+});
+
+test( 'Disabled, optional, readonly, and required are false by default', function( assert ) {
```
notmessenger: Change the name of this test to something more generic, such as "Default values are set to expected values"

tests/unit/mixins/sl-input-based-test.js
```diff
@@ -0,0 +1,200 @@
...
+    );
+
+    assert.strictEqual(
+        subject.get( 'required' ),
+        false,
+        'Required is false by default'
+    );
+});
+
+test( 'Disabled is passed as a parameter', function( assert ) {
```
notmessenger: This test is unnecessary as it is checking Ember functionality and not anything specific to the logic contained within this Mixin.  Any property of any name can be mixed into the object instantiation along with the mixin - this is Ember at work.

tests/unit/mixins/sl-input-based-test.js
```diff
@@ -0,0 +1,200 @@
...
+
+    assert.strictEqual(
+        subject.get( 'required' ),
+        false,
+        'Required is false by default'
+    );
+});
+
+
+test( 'Optional is passed as a parameter', function( assert ) {
```
notmessenger: This test is unnecessary as it is checking Ember functionality and not anything specific to the logic contained within this Mixin.  Any property of any name can be mixed into the object instantiation along with the mixin - this is Ember at work.

tests/unit/mixins/sl-input-based-test.js
```diff
@@ -0,0 +1,200 @@
...
+
+    assert.strictEqual(
+        subject.get( 'required' ),
+        false,
+        'Required is false by default'
+    );
+});
+
+
+test( 'Readonly is passed as a parameter', function( assert ) {
```
notmessenger: This test is unnecessary as it is checking Ember functionality and not anything specific to the logic contained within this Mixin.  Any property of any name can be mixed into the object instantiation along with the mixin - this is Ember at work.

tests/unit/mixins/sl-input-based-test.js
```diff
@@ -0,0 +1,200 @@
...
+    );
+
+    assert.strictEqual(
+        subject.get( 'required' ),
+        false,
+        'Required is false by default'
+    );
+});
+
+test( 'Required is passed as a parameter', function( assert ) {
```
notmessenger: This test is unnecessary as it is checking Ember functionality and not anything specific to the logic contained within this Mixin.  Any property of any name can be mixed into the object instantiation along with the mixin - this is Ember at work.

tests/unit/mixins/sl-input-based-test.js
```diff
@@ -0,0 +1,200 @@
...
+    );
+
+    assert.strictEqual(
+        subject.get( 'readonly' ),
+        false,
+        'Readonly is false by default'
+    );
+});
+
+test( 'All parameters can be passed', function( assert ) {
```
notmessenger: This test is unnecessary as it is checking Ember functionality and not anything specific to the logic contained within this Mixin.  Any property of any name can be mixed into the object instantiation along with the mixin - this is Ember at work.

notmessenger:
Need a test against the DOM that the expected `classnameBindings` values are being bound to the component.  We're not testing Ember's ability to do this binding, but rather that what we are expecting to be bound is being.  This way if someone removes an entry from the `classNameBindings` array the test will fail, meaning that such a change cannot be made without recognition of the impact.

tests/unit/mixins/sl-input-based-test.js
```diff
@@ -0,0 +1,71 @@
...
+    let subject = testObject.create();
+    let classNameBindings = subject.get( 'classNameBindings' );
+
+    assert.equal(
+        classNameBindings.length,
+        expectedClassNameBindings.length,
+        'classNameBindings contains the correct number of elements'
+    );
+
+    classNameBindings.every( ( element, index ) => {
```
notmessenger: Consider the use of `assert.deepEqual()` to reduce required code and to increase speed of the test

tests/unit/mixins/sl-input-based-test.js
```diff
@@ -0,0 +1,60 @@
...
+test( 'Class name bindings contain expected bindings', function( assert ) {
+    let expectedClassNameBindings = [
+        'disabled',
+        'optional',
+        'readonly',
+        'required'
+    ];
+    let testObject = Ember.Object.extend( mixinUnderTest );
+    let subject = testObject.create();
+    let classNameBindings = subject.get( 'classNameBindings' );
```
notmessenger: This assignment is not used anywhere

PR #406, opened 2015-07-09
 Add tests for sl-chart component #207

erangeles:
Hey @notmessenger had to nuke my other PR. I was accidentally pushing my develop branch instead of branching off the develop branch to issue branches. Sorry about this.


addon/components/sl-chart.js
```diff
@@ -77,6 +77,29 @@ export default Ember.Component.extend({
     // Observers
 
     /**
+     * Check passed parameters on initialization
+     *
+     * @function
+     * @throws {ember.assert} Series property must be an array'
```
notmessenger: lone apostrophe at end of line

addon/components/sl-chart.js
```diff
@@ -77,6 +77,29 @@ export default Ember.Component.extend({
...
+     * @throws {ember.assert} Options property must be an Object'
```
notmessenger: lone apostrophe at end of line

notmessenger:
Many lines in tests/unit/components/sl-chart-test.js contain trailing whitespace.  Remove per https://github.com/softlayer/ember-style-guide/blob/master/javascript.md#whitespace

There may be a setting in your editor to automatically do this for you on save.  Which editor do you use?

erangeles:
@notmessenger I'm using sublime at the moment. I'll look into the settings change to do that, I inspected the lines but must have missed that.

notmessenger:
Sublime Text supports this by adding the following line to your preferences file:

"trim_trailing_white_space_on_save": true

There is also a "trim_automatic_white_space": true property

This is also handy - http://stackoverflow.com/questions/10153998/sublime-text-2-view-whitespace-characters

tests/unit/components/sl-chart-test.js
```diff
@@ -0,0 +1,363 @@
...
+        component.set( 'isLoading', true );
+    });
+
+    assert.ok(
+        this.$().hasClass( 'sl-loading' ),
+        'Rendered component gains "sl-loading" class'
+    );
+});
+
+test( 'updateData() is called once', function( assert ) {
```
notmessenger: A more inclusive test would be to check that `updateData()` is called when the `series` property's value changes.  It's true that `updateData()` is called during the initial render of the component but this is only because `series` is initially `null` and the observer defined in `updateData()` is setup to watch `series` and then when the series data passed into the component is set on the `series` property this observer fires again (as an observer watching this property would do).  So it's more about that fact that this observer has been setup correctly and responds to changes to `series` *as well as* that this method is called during the initial rendering of the component.

Additionally, there are no tests for the logic contained within this method.  What about the if/else logic inside the for loop?  Should test that neither of the options in the if/else are called when `series` has not length.  Then test each branch of the if/else.  This method does not return a value so it's not as easy as asserting against a call to it - the side-effects of its execution need to be observed and asserted against.

tests/unit/components/sl-chart-test.js
```diff
@@ -0,0 +1,363 @@
...
+    this.render();
+
+    assert.equal(
+        spy.calledOnce,
+        true,
+        'updateData() is called once'
+    );
+});
+
+test( '"Options" property needs to be an object', function( assert ) {
```
notmessenger: Follow the pattern of https://github.com/softlayer/sl-ember-test-helpers/blob/master/tests/unit/helpers/sl/synchronous/contains-test.js#L15 and comment each type being asserted.

notmessenger: There is a an extra space between the `}  catch(` in each of the blocks below that should be removed.

tests/unit/components/sl-chart-test.js
```diff
@@ -0,0 +1,363 @@
...
+            options: null,
+            series: testSeries
+        });
+    }  catch( error ) {
+        assertionThrown = true;
+    }
+
+    assert.ok(
+        assertionThrown,
+        'testOptions was empty'
```
notmessenger: Change `empty` to `null` - they have different meanings

tests/unit/components/sl-chart-test.js
```diff
@@ -0,0 +1,363 @@
...
+    assert.ok(
+        assertionThrown,
+        'testOptions was a boolean'
+    );
+
+    assertionThrown = false;
+
+    try {
+        this.subject({
+            options: 666,
```
notmessenger: Change to `123`

tests/unit/components/sl-chart-test.js
notmessenger: Also need to test for types `array` and `symbol`

tests/unit/components/sl-chart-test.js
```diff
@@ -0,0 +1,363 @@
...
+            options: testOptions,
+            series: null
+        });
+    }  catch( error ) {
+        assertionThrown = true;
+    }
+
+    assert.ok(
+        assertionThrown,
+        'testSeries was empty'
```
notmessenger: Change `empty` to `null` - they have different meanings

tests/unit/components/sl-chart-test.js
```diff
@@ -0,0 +1,363 @@
...
+        assertionThrown = true;
+    }
+
+    assert.ok(
+        !assertionThrown,
+        'testOptions was an Object'
+    );
+});
+
+test( '"Series" property needs to be an array', function( assert ) {
```
notmessenger: Also need to test for types `object` and `symbol`

notmessenger: There is a an extra space between the `}  catch(` in each of the blocks below that should be removed.

tests/unit/components/sl-chart-test.js
```diff
@@ -0,0 +1,363 @@
...
+        assertionThrown,
+        'testSeries was a boolean'
+    );
+
+    assertionThrown = false;
+
+    try {
+        this.subject({
+            options: testOptions,
+            series: 666
```
notmessenger: Change to `123`

tests/unit/components/sl-chart-test.js
```diff
@@ -0,0 +1,363 @@
...
+    });
+
+    assert.equal( component.get( 'style' ).string, 'height: auto; width: auto;' );
+
+    component.set( 'height', 10 );
+
+    assert.equal( component.get( 'style' ).string, 'height: 10; width: auto;' );
+});
+
+test( "Title property is set", function( assert ) {
```
notmessenger: The test for the `title` property shouldn't be that it can be set on the component and read off of it for two reasons: 1) that is expected Ember functionality that is covered by Ember tests and 2) any property can be set and read this way on a component, so what makes the `title` property special?  Only if we're doing something or reacting to a property that makes it worthy of testing.  In the case of `title` there is no component logic associated with it, but there is template logic, which is what this test should cover.

tests/unit/components/sl-chart-test.js
```diff
@@ -0,0 +1,363 @@
...
+    });
+
+    assert.equal( component.get( 'title' ), null );
+
+    component.set( 'title', 'Peak server hours' );
+
+    assert.equal( component.get( 'title' ), 'Peak server hours' );
+});
+
+test( "Chart property is set after setupChart() is ran", function( assert ) {
```
notmessenger: This should be tested as part of the test of `setupChart()`, which doesn't exist, though I get that this test was part of that attempt.  The logic on lines 167-169 (the last 3 of `setupChart()`) need to have tests around them.  The "updateData() is called once" test can likely be rolled into the same set.

tests/unit/components/sl-chart-test.js
```diff
@@ -0,0 +1,363 @@
...
+    );
+});
+
+test( "Style method returns correct string", function( assert ) {
+    let component = this.subject({
+        options: testOptions, 
+        series: testSeries
+    });
+
+    assert.equal( component.get( 'style' ).string, 'height: auto; width: auto;' );
```
notmessenger: Modify whitespace to conform to https://github.com/softlayer/ember-style-guide#whitespace

tests/unit/components/sl-chart-test.js
```diff
@@ -0,0 +1,363 @@
...
+
+    component.set( 'height', 10 );
+
+    assert.equal( component.get( 'style' ).string, 'height: 10; width: auto;' );
```
notmessenger: Modify whitespace to conform to https://github.com/softlayer/ember-style-guide#whitespace

tests/unit/components/sl-chart-test.js
```diff
@@ -0,0 +1,363 @@
...
+});
+
+test( "Title property is set", function( assert ) {
+    let component = this.subject({
+        options: testOptions, 
+        series: testSeries
+    });
+
+    assert.equal( component.get( 'title' ), null );
```
notmessenger: Modify whitespace to conform to https://github.com/softlayer/ember-style-guide#whitespace

tests/unit/components/sl-chart-test.js
```diff
@@ -0,0 +1,363 @@
...
+
+    component.set( 'title', 'Peak server hours' );
+
+    assert.equal( component.get( 'title' ), 'Peak server hours' );
```
notmessenger: Modify whitespace to conform to https://github.com/softlayer/ember-style-guide#whitespace

tests/unit/components/sl-chart-test.js
```diff
@@ -0,0 +1,363 @@
...
+});
+
+test( "Chart property is set after setupChart() is ran", function( assert ) {
+    let component = this.subject({
+        options: testOptions, 
+        series: testSeries
+    });
+
+    assert.equal( component.get( 'chart' ), null);
```
notmessenger: If were keeping this tests, would need to have modified whitespace to conform to https://github.com/softlayer/ember-style-guide#whitespace

tests/unit/components/sl-chart-test.js
```diff
@@ -0,0 +1,363 @@
...
+
+    this.render();
+
+    assert.ok( component.get( 'chart' ), 'chart is rendered' );
```
notmessenger: If were keeping this tests, would need to have modified whitespace to conform to https://github.com/softlayer/ember-style-guide#whitespace

notmessenger:
I was able to comment on the use of `const` over `let` on those lines that showed up on the diff but hopefully it goes without saying that this should be applied in all scenarios, not just the ones I commented on.

addon/components/sl-chart.js
```diff
@@ -87,18 +110,62 @@ export default Ember.Component.extend({
...
 
-            let options = Ember.$.extend( true, {
-                title: '',
+            return Ember.$.extend( true, {
                 chart: {
                     animation: false,
                     backgroundColor: 'rgba(255, 255, 255, 0)',
                     style: chartStyle
                 },
+                title: null,
```
notmessenger: We talked about adding a comment to this property (which is missing) that indicates why this needs to be set.  Maybe we should consider writing a test against it as well to make sure it doesn't accidentally get removed.

erangeles:
@notmessenger fixed all let variable assignments to use const, added tests for the title property to be null. also in regards to your comment "Reconsider making this a computed property that observes options, as there is nothing watching the result returned from highchartsOptions() in order to react to any changes in the configuration data. Part of the consideration may be to implement such reactions."  I only created the computed property, "highchartsOptions" in order to test unit test it.

notmessenger:
RE: comment https://github.com/softlayer/sl-ember-components/pull/406#issuecomment-121720901

The selection of the structure of data within a component to facilitate testing is not a pattern we want to follow.  The code in the component should be correct and then we figure out how to make it testable.  As such, the options can just be a plain object.  If they need to use `this.get()` to assign values, such as like https://github.com/softlayer/sl-ember-components/blob/develop/addon/components/sl-date-picker.js#L299, then the use of `computed()` is allowed, but we do not need to observe a specific property.

tests/unit/components/sl-chart-test.js
```diff
@@ -0,0 +1,617 @@
...
+    }, component.get( 'options' ) || {} );
+
+    assert.deepEqual(
+        options,
+        component.get( 'highchartsOptions' ),
+        'highchartsOptions returns expected options'
+    );
+});
+
+test('title property is not missing in highchartsOptions and set to null', function( assert ) {
```
notmessenger: Spacing after `(`

addon/components/sl-chart.js
```diff
@@ -87,13 +110,55 @@ export default Ember.Component.extend({
...
+     * Updates the chart's series data
+     *
+     * @function
+     * @returns {undefined}
+     */
+    updateData: Ember.observer(
+        'series',
+        function() {
+            let chart = this.get( 'chart' );
+            let series = this.get( 'series' );
```
notmessenger: Should be `const`

addon/components/sl-chart.js
```diff
@@ -87,13 +110,55 @@ export default Ember.Component.extend({
...
+
+    /**
+     * Options for Highcharts
+     *
+     * @function
+     * @returns {Object}
+     */
+     highchartsOptions: Ember.computed(
+        function() {
             let chartStyle = {
```
notmessenger: Should be `const`

addon/components/sl-chart.js
```diff
@@ -87,13 +110,55 @@ export default Ember.Component.extend({
...
                 fontFamily: '"Benton Sans", "Helvetica Neue", Helvetica, Arial, sans-serif',
                 fontSize: '13px'
             };
 
             let options = Ember.$.extend( true, {
```
notmessenger: Should be `const`

notmessenger:
https://github.com/erangeles/sl-ember-components/blob/issue207/addon/components/sl-chart.js#L227 and https://github.com/erangeles/sl-ember-components/blob/issue207/addon/components/sl-chart.js#L228 should be `const`

tests/unit/components/sl-chart-test.js
```diff
@@ -0,0 +1,617 @@
...
+    assert.ok(
+        spy.calledOnce,
+        'updateData() is called once after series modified'
+    );
+});
+
+test( '"Options" property needs to be an object', function( assert ) {
+
+    // Null
+    let assertionThrown = false;
```
notmessenger: Should be `const`

erangeles:
@notmessenger fixed const assignment.

notmessenger:
Please resolve the merge conflicts

erangeles:
@notmessenger merge conflict fixed.

Issue #433, opened 2015-07-16
"TypeError: select2.matcher is not a function" error being thrown in sl-select demo page


erangeles:
when clicked on the label/input field the select drop down results remain in "searching" mode. Nothing drop down menu items get rendered.

erangeles:
issue #434 address this issue. closing.

notmessenger:
What is the cause of this error?  Because none of the changes in https://github.com/softlayer/sl-ember-components/pull/434 jump out at me as being what fixes this, given that the demo at http://softlayer.github.io/sl-ember-components/#/demos/sl-select is from the release on February 24 and the code being changed in the PR has been effect since then with no changes to the related lines of code but now it's throwing this error?

notmessenger:
@erangeles added an explanation at https://github.com/softlayer/sl-ember-components/pull/434#discussion_r35027336

Issue #389, opened 2015-07-05
Dropdown options are shifted off of the dropdown list element.


notmessenger:
Users @huskerdu and @jkgisme on Slack have reported that the dropdown options are shifted off of the dropdown list element. @huskerdu provided these screen shots: http://cl.ly/image/0Q1j3U0w2t1p and http://cl.ly/image/0C1Q3P2l2D2v.  @jkgisme is using Chrome on OS X and reports that it happens because the page's vertical scroll bar disappears when the dropdown list appears so all the content on the page except the dropdown list moves to the right half a centimeter.  They added `html { overflow-y: scroll; }` to the app.less and it seems to have fixed it.

erangeles:
@notmessenger picking this up

Issue #230, opened 2015-02-04
Add tests for sl-select component


notmessenger:

notmessenger:
https://github.com/softlayer/sl-ember-components/issues/250 may possibly become resolved as well as code is potentially refactored to aid in testing

alxyuu:
@notmessenger taking over this issue from @azizpunjani

alxyuu:
disableSearch only applies to single item select boxes per http://select2.github.io/select2/#documentation

Documentation will be updated to reflect this.

notmessenger:
As https://github.com/softlayer/sl-ember-components/issues/16 has not been resolved yet approach the test against the current functionality which is that both the `required` and `optional` properties are treated independently of each other, like is how similarly approached in the sl-textarea, sl-radio-group, and sl-input components

notmessenger:
Replaced by https://github.com/softlayer/sl-ember-components/issues/456

PR #434, opened 2015-07-16
 Dropdown options are shifted off of the dropdown list element. #389

erangeles:
…out of bounds, fixes shifting of page to the right. @notmessenger alex is going to add testing for this in issue 230 https://huboard.com/softlayer/sl-ember-components/#/issues/56565066 , please merge thanks.

### DIFF
--- a/addon/components/sl-select.js
+++ b/addon/components/sl-select.js
@@ -240,7 +240,7 @@ export default Ember.Component.extend( InputBased, TooltipEnabled, {
                 query: ( query ) => {
                     const content = this.get( 'content' ) || [];
                     const optionLabelPath = this.get( 'optionLabelPath' );
-                    const select2 = this;
+                    const select2 = input.data( 'select2' ).opts;
 
                     query.callback({
                         results: content.reduce( ( results, item ) => {
@@ -265,16 +265,6 @@ export default Ember.Component.extend( InputBased, TooltipEnabled, {
 
             input.on( 'change', () => {
                 this.set( 'value', input.select2( 'val' ) );
-            });
-
-            const originalBodyOverflow = document.body.style.overflow || 'auto';
-
-            input.on( 'select2-open', () => {
-                document.body.style.overflow = 'hidden';
-            });
-
-            input.on( 'select2-close', () => {
-                document.body.style.overflow = originalBodyOverflow;
             });
 
             if ( !this.get( 'multiple' ) ) {
--- a/bower.json
+++ b/bower.json
@@ -12,6 +12,7 @@
     "fontawesome": "~4.3.0",
     "highcharts": "~4.1.5",
     "jquery": "^1.11.1",
+    "jquery-mousewheel": "^3.1.12",
     "loader.js": "ember-cli/loader.js#3.2.0",
     "moment": "~2.10.3",
     "moment-timezone": "~0.4.0",
--- a/index.js
+++ b/index.js
@@ -29,6 +29,11 @@ module.exports = {
         });
 
         app.import({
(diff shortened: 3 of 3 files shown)


addon/components/sl-select.js
```diff
@@ -240,7 +240,7 @@ export default Ember.Component.extend( InputBased, TooltipEnabled, {
                 query: ( query ) => {
                     const content = this.get( 'content' ) || [];
                     const optionLabelPath = this.get( 'optionLabelPath' );
-                    const select2 = this;
+                    const select2 = input.data( 'select2' ).opts;
```
notmessenger: What effect does this change have?  Is it related to https://github.com/softlayer/sl-ember-components/issues/433?

bower.json
```diff
@@ -12,6 +12,7 @@
     "fontawesome": "~4.3.0",
     "highcharts": "~4.1.5",
     "jquery": "^1.11.1",
+    "jquery-mousewheel": "^3.1.12",
```
notmessenger: Why is this needed?

notmessenger:
Setting aside whether or not the jquery-mousewheel dependency is need or not, do you not also need to make an addition to https://github.com/softlayer/sl-ember-components/blob/develop/blueprints/sl-ember-components/index.js?

notmessenger:
How were you able to replicate this problem?  Visiting http://softlayer.github.io/sl-ember-components/#/demos/sl-select using Chrome 43.0.2357.130 (64-bit) on OS X 10.10.4 (14E46) I am unable to recreate it.

erangeles:
@notmessenger are you looking at develop branch? this issue was introduced by https://github.com/softlayer/sl-ember-components/pull/404

addon/components/sl-select.js
erangeles: this previously referenced the options scope of select2, this now references the view, matcher is a function in the options of select2.

bower.json
erangeles: the inclusion of jquery-mousewheel prevents scrolling outside the select menu, which is currently broken.

addon/components/sl-select.js
```diff
@@ -267,16 +267,6 @@ export default Ember.Component.extend( InputBased, TooltipEnabled, {
                 this.set( 'value', input.select2( 'val' ) );
             });
 
-            const originalBodyOverflow = document.body.style.overflow || 'auto';
```
erangeles: The removal of this block of code fixes the shifting issue, where the whole page would shift to one side. https://huboard.com/softlayer/sl-ember-components/#/issues/93135058 referenced in this issue.

erangeles:
@notmessenger https://github.com/softlayer/sl-ember-components/blob/develop/blueprints/sl-ember-components/index.js I honestly don't know if this file has any bearing on injecting dependencies ( or what it really is to be honest). I have removed dependencies in the past from this file on purpose and built the project and it worked fine(to test what this file actually does). I will add the jquery-mousewheel dependency here if needed however I did not find a need for it as it worked fine with my current changes here.

notmessenger:
Issue https://github.com/softlayer/sl-ember-components/issues/230 is no longer a valid issue, so any tests it was going to provide against these code changes needs to be incorporated into this PR.

bower.json
notmessenger: If https://github.com/softlayer/sl-ember-components/pull/434#discussion_r35028305 resolved bug https://github.com/softlayer/sl-ember-components/issues/433 then how is this change related or why is it needed?  Is it adding anything to the fix?  If it is something that has been determined as being desired for the functionality of this component then it should be done in a separate issue.

Issue #426, opened 2015-07-16
Identify listener logic using .on() and .off()


notmessenger:
Identify all occurrences where we use `.on()` calls as well as `.off()` calls to remove them and create issues to remove them.  The gist of the description to use in these issues is below:

The `.off()` calls should only remove the events attached to in `.on()` and not be overly-eager with just blanket `.off()` calls.  The preferred way to approach this is to use namespaces with the `.on()` calls so that they can be specifically targeted in the `.off()` calls and provide isolation between components against events on the same element(s).  The naming pattern of this namespace value still needs to be determined.

PR #435, opened 2015-07-16
Closes softlayer/sl-ember-components#426

alxyuu:


alxyuu:
Tests for sl-input and sl-select will be added in #436 and #437, pending the completion of their respective issues.

tests/unit/components/sl-date-picker-test.js
```diff
@@ -367,22 +367,48 @@ test( 'updateDateRange() - when date outside endDate range we show placeholder t
...
     const component = this.subject();
-    const inputElement = this.$( 'input.date-picker' )[ 0 ];
+    const inputElement = this.$( 'input.date-picker' );
+    const jQueryData = Ember.get( Ember.$, '_data' );
+    const hasDatePickerNamespace = () => {
+        let hasNamespace = false;
+
+        assert.equal(
+            Ember.typeOf(
+                Ember.get( jQueryData( inputElement[0], 'events' ), 'changeDate' )
```
notmessenger: Assign both instances of the usage of `jQueryData( inputElement[0], 'events' )` to a `const` and use it within this test.

tests/unit/components/sl-date-picker-test.js
```diff
@@ -367,22 +367,48 @@ test( 'updateDateRange() - when date outside endDate range we show placeholder t
...
-test( 'unregisterEvents() - listens to correct event', function( assert ) {
+test( 'changeDate listener is added and removed from the correct namespace', function( assert ) {
     const component = this.subject();
-    const inputElement = this.$( 'input.date-picker' )[ 0 ];
+    const inputElement = this.$( 'input.date-picker' );
+    const jQueryData = Ember.get( Ember.$, '_data' );
+    const hasDatePickerNamespace = () => {
+        let hasNamespace = false;
+
+        assert.equal(
```
notmessenger: This assertion should not be within this function but instead be stand-alone.

tests/unit/components/sl-date-picker-test.js
```diff
@@ -367,22 +367,48 @@ test( 'updateDateRange() - when date outside endDate range we show placeholder t
...
+            'Datepicker has changeDate listeners'
+        );
 
-    assert.equal(
-        Ember.typeOf( $._data( inputElement, 'events' ).changeDate ),
-        'array',
-        'Datepicker has a changeDate event listener after render'
+        Ember.get( jQueryData( inputElement[0], 'events' ), 'changeDate' ).every(
+            ( element ) => {
+                if ( element.namespace === "sl-date-picker" ) {
```
notmessenger: Use single quotes per https://github.com/softlayer/ember-style-guide/blob/master/javascript.md#strings-1

notmessenger: Change format of comparison per https://github.com/softlayer/ember-style-guide/issues/18

tests/unit/components/sl-date-picker-test.js
```diff
@@ -367,22 +367,48 @@ test( 'updateDateRange() - when date outside endDate range we show placeholder t
...
+        return hasNamespace;
+    };
+
+    assert.ok(
+        hasDatePickerNamespace(),
+        'Datepicker has a changeDate event listener in the correct namespace after render'
     );
 
-    component.trigger( 'willClearRender' );
+    inputElement.on( 'changeDate', function(){} );
```
notmessenger: Why is this code needed?

notmessenger:
Has it been determined that similar changes are not needed for the `sl-menu` and `sl-radio-group` components?

notmessenger:
The "setupDatepicker() - listens to correct event" test needs modification.

tests/unit/components/sl-date-picker-test.js
alxyuu: it's to add a changeDate listener not in our namespace, so we can verify that willClearRender does not remove it

alxyuu:
Oops, missed sl-radio-group since it wasn't using on. The on and off in sl-menu is from ember rather than jquery, which does not support event namespacing.

tests/unit/components/sl-date-picker-test.js
```diff
@@ -367,22 +367,45 @@ test( 'updateDateRange() - when date outside endDate range we show placeholder t
     );
 });
 
-test( 'unregisterEvents() - listens to correct event', function( assert ) {
+test( 'changeDate listener is added and removed from the correct namespace', function( assert ) {
     const component = this.subject();
-    const inputElement = this.$( 'input.date-picker' )[ 0 ];
+    const inputElement = this.$( 'input.date-picker' );
+    const jQueryData = Ember.get( Ember.$, '_data' );
+    const eventData = Ember.get( jQueryData( inputElement[0], 'events' ), 'changeDate' );
```
notmessenger: Can't this data be gotten off of `events` via a `.( 'data' )` call instead of using a private, undocumented, property?

Rather than use this approach at all it is preferred to spy on the jQuery `.on()` (or mock if necessary, similar to the mocks in https://github.com/softlayer/sl-ember-components/blob/develop/tests/unit/mixins/sl-tooltip-enabled-test.js#L6-L24) so that we can see that calls with the expected data are made rather than relying on the API to remain unchanged in jQuery, etc.  The point being, what are we after?  We want to make sure that `.on()` is called with a name spaced value.  We are not looking to test the the jQuery implementation of how it happens, whether the jQuery library code is working, etc.

tests/unit/components/sl-date-range-picker-test.js
```diff
@@ -95,30 +95,45 @@ test( 'Latest start date is the based on max date and end date', function( asser
...
-test( 'Events from start date input are removed upon willClearRender', function( assert ) {
+test( 'changeDate listener is added and removed from the correct namespace', function( assert ) {
     const component = this.subject();
-
-    this.render();
-
-    const startDateInput = this.$( '.sl-daterange-start-date input' )[ 0 ];
+    const inputElement = this.$( '.sl-daterange-start-date input' );
     const jQueryData = Ember.get( Ember.$, '_data' );
+    const eventData = Ember.get( jQueryData( inputElement[0], 'events' ), 'changeDate' );
```
notmessenger: Same comment left on similar logic in sl-date-picker-test.js file applies here as well.

tests/unit/components/sl-grid-test.js
```diff
@@ -334,6 +334,68 @@ test( 'Window resize triggers updateHeight() with "auto" width', function( asser
     );
 });
 
+test( 'Grid adds and removes events from the correct namespace', function( assert ) {
+    const component = this.subject({ columns, content, height: 'auto' });
+    const contentElement = this.$( '.list-pane .content' );
+    const windowElement = Ember.$( window );
+    const jQueryData = Ember.get( Ember.$, '_data' );
```
notmessenger: Same comment left on similar logic in sl-date-picker-test.js file applies here as well.

addon/components/sl-grid.js
```diff
@@ -620,7 +634,7 @@ export default Ember.Component.extend({
      * @returns {undefined}
      */
     disableContinuousPaging() {
-        this.$( '.list-pane .content' ).unbind( 'scroll' );
+        this.$( '.list-pane .content' ).off( 'scroll.sl-grid' );
```
notmessenger: Should this also be elementId specific?

addon/components/sl-grid.js
```diff
@@ -630,7 +644,7 @@ export default Ember.Component.extend({
      * @returns {undefined}
      */
     enableContinuousPaging() {
-        this.$( '.list-pane .content' ).bind( 'scroll', ( event ) => {
+        this.$( '.list-pane .content' ).on( 'scroll.sl-grid', ( event ) => {
```
notmessenger: Should this also be elementId specific?

addon/components/sl-grid.js
alxyuu: No, because it's already specific to this component via this.$
The window listener needed the element id since multiple grids could add resize listeners to the window.

tests/unit/components/sl-radio-group-test.js
```diff
@@ -145,3 +145,50 @@ test( "Inline false sets sl-radio children's inline property to false", function
...
+test( 'change listener is added and removed from the correct namespace', function( assert ) {
+    const component = this.subject({
+        name: 'test',
+        template: Ember.Handlebars.compile(
+            '{{sl-radio label="One" value="one"}}'
+        )
+    });
+    const inputElement = this.$( 'input[name=test]:radio' );
+    const jQueryData = Ember.get( Ember.$, '_data' );
+    const eventData = Ember.get( jQueryData( inputElement[0], 'events' ), 'change' );
```
notmessenger: Same comment left on similar logic in sl-date-picker-test.js file applies here as well.

notmessenger:
https://github.com/softlayer/sl-ember-components/issues/426 has been re-prioritized.

Issue #423, opened 2015-07-15
Add test to sl-date-range-picker for helpText


alxyuu:
There should be a test that verifies that helpText is undefined by default and that no dom element is created, as well as verify the correct element is created when helpText is set

notmessenger:
https://github.com/softlayer/sl-ember-components/pull/412 will address the test about the correct element being created, so this issue only needs to test that `helpText` is passed along.

alxyuu:
Actually, helpText isn't being passed along, it's an element in the date range picker. I missed testing for it in #210

notmessenger:
Replaced by https://github.com/softlayer/sl-ember-components/issues/455

PR #442, opened 2015-07-17
Closes softlayer/sl-ember-components#423

juwara0:
I fixed issues: 423 and 424 in this work. I also updated the demo app and docBlocks for the date-picker and date range picker so that startDate/endDate are consistent and correct.

### DIFF
--- a/addon/components/sl-date-range-picker.js
+++ b/addon/components/sl-date-range-picker.js
@@ -54,7 +54,7 @@ export default Ember.Component.extend({
     /**
      * The help text below the date range picker
      *
-     * @type {String}
+     * @type {?String}
      */
     helpText: null,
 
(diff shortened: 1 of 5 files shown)


addon/components/sl-date-range-picker.js
```diff
@@ -45,6 +52,13 @@ export default Ember.Component.extend({
     format: 'mm/dd/yyyy',
 
     /**
+     * The help text below the date range picker
+     *
+     * @type {String}
```
notmessenger: Can also be null

juwara0: I have updated this and label below to indicate null is allowed as well.

notmessenger:
https://github.com/softlayer/sl-ember-components/issues/423 has been re-prioritized

PR #438, opened 2015-07-16
Closes softlayer/sl-ember-components#61

alxyuu:

PR #443, opened 2015-07-17
Closes softlayer/sl-ember-component#230

alxyuu:


addon/components/sl-select.js
```diff
@@ -103,7 +103,24 @@ export default Ember.Component.extend( InputBased, TooltipEnabled, {
     destroySelect2: Ember.on(
         'willClearRender',
         function() {
-            this.input.off( 'change' ).select2( 'destroy' );
+            this.get( 'input' ).off( 'change' ).select2( 'destroy' );
```
notmessenger: Doesn't need namespace referencing?  Know you are doing that in a separate PR but are the merges going to be clean between it and this one?

addon/components/sl-select.js
```diff
@@ -116,20 +133,71 @@ export default Ember.Component.extend( InputBased, TooltipEnabled, {
...
-                    }
+    /**
+     * Select2 fomartResult function
+     *
+     * @function
+     * @returns {Function}
+     */
+    select2FormatResult: Ember.computed(
+        function() {
+            return ( item ) => {
```
notmessenger: Original logic returned `null` if no item value was passed.

notmessenger: Though this will likely be okay upon further review of the logic.

addon/components/sl-select.js
```diff
@@ -141,58 +209,83 @@ export default Ember.Component.extend( InputBased, TooltipEnabled, {
...
+     * @function
+     * @returns {Function}
+     */
+    select2InitSelection: Ember.computed(
+        function() {
+            return ( element, callback ) => {
+                const value = element.val();
 
+                if ( !value || !value.length ) {
+                    callback( [] );
```
notmessenger: Can you help me understand the removal of the `return` statement?

addon/components/sl-select.js
```diff
@@ -141,58 +209,83 @@ export default Ember.Component.extend( InputBased, TooltipEnabled, {
...
+                } else {
                     const content = this.get( 'content' );
                     const contentLength = content.length;
                     const filteredContent = [];
                     const multiple = this.get( 'multiple' );
                     const optionValuePath = this.get( 'optionValuePath' );
-                    const values = value.split( ',' );
+                    const values = multiple ? value.split( ',' ) : [ value ];
```
notmessenger: This change is not necessary.  The use of `split` against a single string value will result in the creation of an array containing the original string.

addon/components/sl-select.js
```diff
@@ -228,67 +321,48 @@ export default Ember.Component.extend( InputBased, TooltipEnabled, {
...
-                            }
-
-                            return results;
-                        }, [] )
-                    });
+                    if ( filteredContent.length > 0 ) {
+                        callback( multiple ? filteredContent : filteredContent[0] );
+                    } else {
+                        callback( [] );
+                    }
```
notmessenger: I do not understand the benefit of this change over the original code.

addon/components/sl-select.js
```diff
@@ -116,20 +133,71 @@ export default Ember.Component.extend( InputBased, TooltipEnabled, {
...
+
+            input.on( 'select2-open', () => {
+                document.body.style.overflow = 'hidden';
+            });
+
+            input.on( 'select2-close', () => {
+                document.body.style.overflow = originalBodyOverflow;
+            });
+
+            this.$( 'input.select2-input' ).attr(
```
notmessenger: Is this call even necessary with a/the `placeholder` property being passed to the select2 initializer as an option?

addon/components/sl-select.js
```diff
@@ -228,67 +321,48 @@ export default Ember.Component.extend( InputBased, TooltipEnabled, {
...
+     *
+     * @function
+     * @returns {Function}
+     */
+    select2Query: Ember.computed(
+        function() {
+            return ( query ) => {
+                const content = this.get( 'content' );
+                const optionLabelPath = this.get( 'optionLabelPath' );
+                const select2 = this.get( 'input' ).data( 'select2' ).opts;
```
notmessenger: What is this change providing?

addon/components/sl-select.js
alxyuu: Yes, this sets the placeholder for the search box in single select boxes.

addon/components/sl-select.js
alxyuu: according to the select2 documentation, this function shouldn't be returning anything, it should just be using the callback. Not that returning would break anything, but it was unnecessary, since the callback doesn't return anything either.

addon/components/sl-select.js
alxyuu: My thought process went something like if the user wanted an option with a comma in it. It wouldn't make sense if a user had a single select and passed value="123,123", and actually had an option "123,123". By splitting on the comma the selection would end up disabled since it can't find 123 in the options.

addon/components/sl-select.js
alxyuu: Really just personal sanity. If there was nothing matched for a single select, filteredContent[0] would return undefined. This way we can consistently return an a type of {Array|String}

addon/components/sl-select.js
alxyuu: Like Eric mentioned in #434, it's due to a change in scope from #404 that was not fixed. I decided to just add the fix here instead of waiting to merge.

alxyuu:
@notmessenger I'd like to hold this back pending the merge of #434, #435, and #438 since there will be merge conflicts

notmessenger:
https://github.com/softlayer/sl-ember-components/issues/230 has been re-prioritized

PR #476, opened 2015-07-24
Remove alias to _parentView._controller

erangeles:
initial commit, wanted to get your input on it @notmessenger . I reference this source http://discuss.emberjs.com/t/sub-component-action-not-bubbling-up/6739 for this fix.


erangeles:
initial commit, wanted to get your input on it @notmessenger . I reference this source http://discuss.emberjs.com/t/sub-component-action-not-bubbling-up/6739 for this fix.

notmessenger:
Would like your input on this @joshforisha

joshforisha:
Part of the reason for the alias initially was so that the object the rowActions were run on could be customized. I suppose the same thing can be done by setting the targetAction on the component, but that will also affect other states of the grid unrelated to the rowActions.

If you don't think that changing the rowActions context is necessary, then this change would be fine.

notmessenger:
Does yielded context give us a way forward on this @erangeles, as mentioned in the issue?

erangeles:
@notmessenger yes it does.

notmessenger:
I would prefer this approach then as it is more "standard" and less likely to change on us.  Your thoughts @joshforisha?

joshforisha:
Sounds good.

notmessenger:
See notes on issue

PR #484, opened 2015-07-27
removing sl-datepicker from browsers dependancy grid

SpikedKira:
first :dango:


tests/dummy/app/controllers/browsers.js
```diff
@@ -30,9 +30,6 @@ export default Ember.Controller.extend({
             name: 'drop-button',
             lib: { 'twb': true }
         }, {
-            name: 'drop-option',
```
notmessenger: This is not related to the `sl-panel` component.  Is this modification intentional?

tests/dummy/app/controllers/browsers.js
```diff
@@ -47,12 +44,9 @@ export default Ember.Controller.extend({
         }, {
             name: 'modal',
             lib: { 'twb': true }
-        }, {
-            name: 'pagination',
```
notmessenger: This is not related to the `sl-panel` component.  Is this modification intentional?

tests/dummy/app/controllers/browsers.js
```diff
@@ -69,9 +63,6 @@ export default Ember.Controller.extend({
             name: 'span',
             lib: { 'sec': true }
         }, {
-            name: 'tab-pane',
```
notmessenger: This is not related to the `sl-panel` component.  Is this modification intentional?

notmessenger:
@SpikedKira is going to re-submit this PR

PR #496, opened 2015-07-29
Added vimrc to git ignore

azizpunjani:
So that projects can have their own .vimrc files with project specific settings without getting accidentally checked in.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@
 npm-debug.log
 testem.log
 *.swp
+.vimrc


notmessenger:
These are files specific to an individual's development environment and not anything related to this repository.  It is not the repository's responsibility to accommodate all of these possible scenarios.  The `core.excludesFile` option described at http://git-scm.com/docs/gitignore seems well suited for this scenario.

Issue #497, opened 2015-07-31
Refactor document.activeElement checks


azizpunjani:
To work around the document having focus when focusing an element, we can use the native DOM `dispatchEvent` which does not require the document to be focused. The code outlined does an `Ember.run.later` and checks `document.activeElement`. This can now be refactored to wire up the the `focus` event and trigger the `focus` event via the code.

```
const evt = new window.Event( 'focus' );
input.get( 0 ).dispatchEvent( evt );
```
https://github.com/softlayer/sl-ember-components/blob/d732f7fb8b2383e76ef92245e335d21781402d3e/tests/unit/components/sl-date-range-picker-test.js#L39

notmessenger:
Resolved by https://github.com/softlayer/sl-ember-components/pull/538

PR #528, opened 2015-08-04
 Refactor document.activeElement checks #497

erangeles:


notmessenger:
The changes made to addon/components/sl-date-range-picker.js need to be in a separate issue and PR, as they are unrelated to https://github.com/softlayer/sl-ember-components/issues/497

Issue #452, opened 2015-07-21
Upgrade to Ember CLI 1.13.x and Ember 1.13.x


joshforisha:

notmessenger:
Resolved by https://github.com/softlayer/sl-ember-components/pull/543

PR #542, opened 2015-08-11
Closes softlayer/sl-ember-components#452

notmessenger:


notmessenger:
Opened against the incorrect branch

PR #662, opened 2015-08-13
Fix double render

azizpunjani:
Since sl-radio `input` is within a label, a `for` attribute is not required according to the spec.

### DIFF
--- a/addon/components/sl-select.js
+++ b/addon/components/sl-select.js
@@ -43,13 +43,6 @@ export default Ember.Component.extend( InputBased, TooltipEnabled, {
     disableSearch: false,
 
     /**
-     * Unique input id that will get generated and set on init of component
-     *
-     * @type {?String}
-     */
-    inputId: null,
-
-    /**
      * The internal input element, used for Select2's bindings
      *
      * @type {?Object}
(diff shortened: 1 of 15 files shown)


addon/components/sl-select.js
```diff
@@ -43,6 +43,13 @@ export default Ember.Component.extend( InputBased, TooltipEnabled, {
     disableSearch: false,
 
     /**
+     * Unique input id that will generated and set on init of component
```
azizpunjani: Need to fix typo here.

addon/components/sl-textarea.js
```diff
@@ -110,6 +110,13 @@ export default Ember.Component.extend( InputBased, TooltipEnabled, {
     spellcheck: Spellcheck.Direction,
 
     /**
+     * Unique textarea id that will generated and set on init of component
```
azizpunjani: Fix typo

addon/templates/components/sl-select.hbs
```diff
@@ -13,7 +13,7 @@
 {{/if}}
 
 {{input
-    viewName="sl-select-view"
+    id=inputId
```
notmessenger: Thinking "inputId" should at least be "selectId" to keep with the naming convention used by `sl-textarea`, though what about using something more generic instead?

Going with the generic, 2 out of 3 of these components need this functionality, which is identical, so what about moving it into the `sl-input-based` mixin?

azizpunjani: Using inputId consistently now for both `sl-textarea` and `sl-select`

azizpunjani:
@notmessenger Moved inputId generation to `sl-input-based`.

notmessenger:
Research use of elementId

azizpunjani:
@notmessenger did some research, this is best approach we have. I initially thought we could use elementId + '-input' in the view itself but handlebars doesn't like that, it's better to put it in a function like we have above.

addon/templates/components/sl-select.hbs
notmessenger: RESOLVED

Issue #510, opened 2015-08-04
Computed properties should return null value by default


notmessenger:

notmessenger:
This issue needs to be merged into `master` before work on https://github.com/softlayer/sl-ember-components/issues/583 begins.  If it cannot be then work on it needs to be delayed until the release of v0.10.0

PR #665, opened 2015-08-14
Closes softlayer/sl-ember-components#510

theoshu:
edited sl-date-time.js so computed properties return null

### DIFF
--- a/addon/components/sl-date-time.js
+++ b/addon/components/sl-date-time.js
@@ -187,7 +187,7 @@ export default Ember.Component.extend( TooltipEnabled, {
         'momentValue',
         function() {
             const momentValue = this.get( 'momentValue' ),
-                  timezone = this.get( 'timezone' );
+                timezone = this.get( 'timezone' );
 
             if ( momentValue && timezone ) {
                 return momentValue
--- a/addon/components/sl-date-time.js
+++ b/addon/components/sl-date-time.js
@@ -106,14 +106,13 @@ export default Ember.Component.extend( TooltipEnabled, {
         'value',
         function() {
             const timezoneString = this.get( 'timezoneString' );
-            const value = this.get( 'value' ) ? this.get( 'value' ) : new Date();
 
             if ( timezoneString ) {
-                return window.moment.tz( value, timezoneString )
+                return window.moment.tz( this.get( 'value' ), timezoneString )
                     .format( 'YYYY-MM-DD HH:mm z' );
             }
 
-            return value.format( 'YYYY-MM-DD HH:mm Z' );
+            return null;
         }
     ),
 
@@ -168,7 +167,8 @@ export default Ember.Component.extend( TooltipEnabled, {
             const locale = this.get( 'locale' );
 
             if ( locale ) {
-                return window.moment( this.get( 'value' ) ).locale( locale );
+                return window.moment( this.get( 'value' ) )
+                    .locale( locale );
             }
 
             return null;
@@ -185,11 +185,12 @@ export default Ember.Component.extend( TooltipEnabled, {
         'timezone',
         'momentValue',
         function() {
-            const momentValue = this.get( 'momentValue' );
-            const timezone = this.get( 'timezone' );
+            const momentValue = this.get( 'momentValue' ),
+                timezone = this.get( 'timezone' );
 
             if ( momentValue && timezone ) {
-                return momentValue.tz( timezone ).format( 'z' );
+                return momentValue
+                    .tz( timezone ).format( 'z' );
             }
 
             return null;
--- a/addon/components/sl-date-time.js
+++ b/addon/components/sl-date-time.js
@@ -105,9 +105,15 @@ export default Ember.Component.extend( TooltipEnabled, {
         'timezoneString',
         'value',
         function() {
-            return window.moment( this.get( 'value' ) )
-                .format( 'YYYY-MM-DD HH:mm ' ) +
-                this.get( 'timezoneString' );
+            const timezoneString = this.get( 'timezoneString' );
+
+            if ( timezoneString ) {
+                return window.moment( this.get( 'value' ) )
+                    .format( 'YYYY-MM-DD HH:mm ' ) +
+                    timezoneString;
+            }
+
+            return null;
         }
     ),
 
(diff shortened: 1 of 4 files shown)


notmessenger:
@theoshu The build is failing.  Can select the "Details" link to view the TravisCI build.  The `EMBER_TRY_SCENARIO=default` job is the one to select to expand to see what errors have been encountered.

notmessenger:
@theoshu Please merge the `master` branch into this one and run the tests against it in your development environment.  I have merged https://github.com/softlayer/sl-ember-components/pull/672 into the `master` branch which should provide the exact same PhantomJS installation into the dev and TravisCI environments.

theoshu:
I ran > git pull origin master and am up to date (???)

notmessenger:
@theoshu If it showed that files were updated then sounds like you are good.  Push those changes up to this PR so can see how TravisCI handles them vs running the tests in your local dev.  My effort is/was to resolve the scenario where tests were passing in your PR when checked out and ran locally versus in TravisCI where they were failing.

theoshu:
Turns out the test was wrong. We were testing with central time zone, which would never fail on my local copy.

addon/components/sl-date-time.js
```diff
@@ -105,9 +105,14 @@ export default Ember.Component.extend( TooltipEnabled, {
         'timezoneString',
         'value',
         function() {
-            return window.moment( this.get( 'value' ) )
-                .format( 'YYYY-MM-DD HH:mm ' ) +
-                this.get( 'timezoneString' );
+            const timezoneString = this.get( 'timezoneString' );
+
+            if ( timezoneString ) {
```
notmessenger: Should `value` also be evaluated, like in `timezoneString()`?

addon/components/sl-date-time.js
```diff
@@ -169,8 +185,15 @@ export default Ember.Component.extend( TooltipEnabled, {
         'timezone',
         'momentValue',
         function() {
-            return this.get( 'momentValue' )
-                .tz( this.get( 'timezone' ) ).format( 'z' );
+            const momentValue = this.get( 'momentValue' ),
+                timezone = this.get( 'timezone' );
```
notmessenger: https://github.com/softlayer/ember-style-guide/blob/master/javascript.md#variables

addon/components/sl-date-time.js
```diff
@@ -169,8 +185,15 @@ export default Ember.Component.extend( TooltipEnabled, {
...
+
+            if ( momentValue && timezone ) {
+                return momentValue
```
notmessenger: Either place all of these calls on the same line (preferred) or put each new method call on its own line.

tests/unit/components/sl-date-time-test.js
```diff
@@ -16,15 +16,28 @@ test( 'Default class names are present', function( assert ) {
 
 test( 'Attribute "datetime" is properly set', function( assert ) {
     this.subject({
-        timezone: 'America/Chicago',
-        value: new Date( 2015, 0, 1 )
+        format: 'datetime',
```
notmessenger: This is the default value - is there value in specifying it in this assertion?

tests/unit/components/sl-date-time-test.js
```diff
@@ -16,15 +16,28 @@ test( 'Default class names are present', function( assert ) {
...
 
     assert.equal(
         this.$().attr( 'datetime' ),
-        '2015-01-01 00:00 CST',
+        '2015-01-01 00:00 MST',
         '"datetime" attribute is expected value'
     );
+
+    this.subject({
+        format: 'datetime',
```
notmessenger: This is the default value - is there value in specifying it in this assertion?

tests/unit/components/sl-date-time-test.js
```diff
@@ -16,15 +16,28 @@ test( 'Default class names are present', function( assert ) {
 
 test( 'Attribute "datetime" is properly set', function( assert ) {
     this.subject({
-        timezone: 'America/Chicago',
-        value: new Date( 2015, 0, 1 )
+        format: 'datetime',
+        timezone: 'America/Denver',
```
notmessenger: I'm assuming that this change is related to your comment on the PR about how the test was incorrect because of the timezone used and that it was not failing in your dev environment?  If so, this is not an acceptable fix.  All this change does is "fix" it for any devs in the `America/Chicago` timezone but has moved the problem to those in the `America/Denver` timezone.  The approach of this test needs to be re-visited.

theoshu: The real fix to the test is on line 21 (replaces line 20 of original). Instantiating a datetime via new Date ( 2015, 0, 1) is the variable as it uses the browser timezone and comparing the values will only pass if the test is run on a machine with that specified timezone. It won't break in MST.

notmessenger: If line 21 is the real fix, then why the other changes?  Specifically, why the timezone change?

theoshu: I was specifically trying to get the test to fail. Apparently the github server where travis does the testing is in EDT (currently), so that's why the America/Chicago failed, but not locally. America/Denver isn't CDT or EDT, so it failing becomes more likely. Timezones make my head hurt.

addon/components/sl-date-time.js
```diff
@@ -169,8 +185,14 @@ export default Ember.Component.extend( TooltipEnabled, {
...
-            return this.get( 'momentValue' )
-                .tz( this.get( 'timezone' ) ).format( 'z' );
+            const momentValue = this.get( 'momentValue' );
+            const timezone = this.get( 'timezone' );
+
+            if ( momentValue && timezone ) {
+                return momentValue.tz( timezone ).format( 'z' );
+            }
+
+            return null;
```
notmessenger: I have created https://github.com/softlayer/sl-ember-components/issues/701 to address a deficiency with this component.  With this issue in place `this.get( 'timezone' )` will also return a correct value meaning that this computed property will also always return a value, negating the need to modify the code to return `'null`.  Your changes can be removed.

addon/components/sl-date-time.js
```diff
@@ -154,8 +165,13 @@ export default Ember.Component.extend( TooltipEnabled, {
...
         function() {
-            return window.moment( this.get( 'value' ) )
-                .locale( this.get( 'locale' ) );
+            const locale = this.get( 'locale' );
+
+            if ( locale ) {
+                return window.moment( this.get( 'value' ) ).locale( locale );
+            }
+
+            return null;
```
notmessenger: `locale` (https://github.com/theoshu/sl-ember-components/blob/issue510/addon/components/sl-date-time.js#L67) and `value` (https://github.com/theoshu/sl-ember-components/blob/issue510/addon/components/sl-date-time.js#L90) are always set so this computed property always returns a value, negating the need to return a null.  These changes can be reverted.

addon/components/sl-date-time.js
```diff
@@ -105,9 +105,15 @@ export default Ember.Component.extend( TooltipEnabled, {
...
-                this.get( 'timezoneString' );
+            const timezoneString = this.get( 'timezoneString' );
+            const value = this.get( 'value' ) ? this.get( 'value' ) : new Date();
+
+            if ( timezoneString ) {
+                return window.moment.tz( value, timezoneString )
+                    .format( 'YYYY-MM-DD HH:mm z' );
+            }
+
+            return value.format( 'YYYY-MM-DD HH:mm Z' );
```
notmessenger: Due to `value` (https://github.com/theoshu/sl-ember-components/blob/issue510/addon/components/sl-date-time.js#L90) defaulting to `new Date()` and the comments made about `timezoneString` in this same PR this computed property will always return a value, negating the need for any of these changes.

addon/components/sl-date-time.js
```diff
@@ -122,23 +128,28 @@ export default Ember.Component.extend( TooltipEnabled, {
...
+                    default:
+                    case Format.DATETIME:
+                        if ( this.get( 'timezoneString' ) ) {
+                            formattedString =
+                                momentValue.format( 'dddd, MMMM Do YYYY, h:mm A' ) +
+                                ' ' +
+                                this.get( 'timezoneString', 'en' );
+                        }
+                        break;
+                }
```
notmessenger: It is acceptable for this computed property to return an empty string value, so all of these changes can be reverted.

notmessenger:
The end result of all of the comments I have made against the addon/components/sl-date-time.js file is that no changes are necessary.  Given this, are any necessary for tests/unit/components/sl-date-time-test.js?  Even if they are not, because they were directly related to the changes that are no longer needed, I remain interested in discussing the timezone issues you were having and whether this condition still exists within the test(s).

PR #676, opened 2015-08-20
Closes softlayer/sl-ember-components#512

theoshu:
added return null to sl-grid computed properties

### DIFF
--- a/addon/components/sl-grid.js
+++ b/addon/components/sl-grid.js
@@ -485,6 +485,70 @@ export default Ember.Component.extend({
     ),
 
     /**
+     * Update the panes' heights according to `height` property value
+     *
+     * @function
+     * @returns {undefined}
+     */
+    updateHeight: Ember.on(
+        'didInsertElement',
+        function() {
+            if ( !this.$() ) {
+                return;
+            }
+
+            const componentHeight = this.get( 'height' );
+            const gridHeader = this.$( '.grid-header' );
+            const detailHeader = this.$( '.detail-pane header' );
+            const detailFooter = this.$( '.detail-pane footer' );
+            const listHeader = this.$( '.list-pane .column-headers' );
+            const listFooter = this.$( '.list-pane footer' );
+
+            const detailHeaderHeight = detailHeader ?
+                parseInt( detailHeader.css( 'height' ) ) : 0;
+
+            const detailFooterHeight = detailFooter ?
+                parseInt( detailFooter.css( 'height' ) ) : 0;
+
+            const gridHeaderHeight = gridHeader ?
+                parseInt( gridHeader.css( 'height' ) ) : 0;
+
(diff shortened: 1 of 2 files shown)


addon/components/sl-grid.js
```diff
@@ -665,7 +669,14 @@ export default Ember.Component.extend({
         'content.length',
         'totalCount',
         function() {
```
notmessenger: `@returns` value needs updated

notmessenger:
Should `totalPages()` be modified?

notmessenger:
No tests needed to be modified?

addon/components/sl-grid.js
notmessenger: Does this computed property need to return `null`?  Seems `Boolean` is the only valid return value?

theoshu:
@notmessenger re: your question about `totalPages()`, are you asking why did I change it?

notmessenger:
@theoshu Regarding your question about my `totalPages()` comment: No, my comment was regarding that there we no changes made to `totalPages()` and perhaps there needs to be?  Upon further review however, specifically line https://github.com/theoshu/sl-ember-components/blob/issue512/addon/components/sl-grid.js#L499, if the value of `totalPages()` is ever anything undesired, this line will ensure correct behavior of the component.

addon/components/sl-grid.js
theoshu: fixed

addon/components/sl-grid.js
```diff
@@ -546,6 +548,8 @@ export default Ember.Component.extend({
                     this.get( 'totalCount' ) / this.get( 'pageSize' )
                 );
             }
+
+            return null;
```
notmessenger: Change DocBlock `@returns {Number|undefined}` to `@returns {?Number}`

notmessenger:
Why is https://github.com/theoshu/sl-ember-components/blob/issue512/tests/unit/components/sl-grid-test.js#L323 and https://github.com/theoshu/sl-ember-components/blob/issue512/tests/unit/components/sl-grid-test.js#L335 being set to the same value twice?  Doesn't seem like https://github.com/theoshu/sl-ember-components/blob/issue512/tests/unit/components/sl-grid-test.js#L326 and https://github.com/theoshu/sl-ember-components/blob/issue512/tests/unit/components/sl-grid-test.js#L338 are needed twice either.

addon/components/sl-grid.js
theoshu: Done

notmessenger:
`showPagination()`, 'sortedColumn()`, and `totalPages()` should be in the "Methods" section.

addon/components/sl-grid.js
notmessenger: RESOLVED

PR #677, opened 2015-08-20
Closes softlayer/sl-ember-components#514

theoshu:
This adds checks for currentPage and totalPages to sl-pagination's computed properties and returns null if they do not exist.


notmessenger:
Neither of these changes are required

PR #691, opened 2015-08-25
Closes softlayer/sl-ember-components#issue518

theoshu:
Added return null to sl-drop-button.js


notmessenger:
`@returns` type of `themeClass()` needs to reflect possible `null` return value

theoshu:
huh?

notmessenger:
https://github.com/theoshu/sl-ember-components/blob/issue518/addon/components/sl-drop-button.js#L142

Instead of `@returns {String}` is should be `@returns {?String}`, where the `?` denotes a nullable type (http://usejsdoc.org/tags-type.html)

PR #702, opened 2015-08-26
Issue523

theoshu:
Added test to check if passing undefined reverts back to default. No changes necessary to the component.

### DIFF
--- a/tests/unit/components/sl-date-picker-test.js
+++ b/tests/unit/components/sl-date-picker-test.js
@@ -418,5 +418,50 @@ test( 'options() - object contains the expected keys', function( assert ) {
         Object.keys( optionsMockObject ).sort(),
         'The options object contains the expected keys'
     );
+
+    const emptyComponent = this.subject({
+        autoclose: undefined,
+        calendarWeeks: undefined,
+        clearBtn: undefined,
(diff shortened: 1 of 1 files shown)


tests/unit/components/sl-date-picker-test.js
```diff
@@ -418,5 +418,50 @@ test( 'options() - object contains the expected keys', function( assert ) {
...
+            language: 'en',
+            multidate: false,
+            orientation: 'auto',
+            startDate: null,
+            startView: 'month',
+            todayBtn: false,
+            todayHighlight: false,
+            weekStart: 0
+        },
+        'Undefined properties change to default values'
```
notmessenger: If you're wanting to test this wouldn't a better way to be to just call `this.subject()` without any options?  That is what the template usage would be: ``{{sl-date-picker}}`.  You wouldn't be doing `autoclose = undefined;` and then `{{sl-date-picker autoclose=autoclose}}`

theoshu: This change might be completely unnecessary. I was worried that if they are explicitly passed undefined that it would fall down go boom, but it doesn't. If unnecessary, this can probably be closed.

Issue #727, opened 2015-09-01
Refactor integration tests


notmessenger:
Replace use of `.sl-span` per https://github.com/softlayer/ember-style-guide/issues/29, as well as utilize the same referenced selector in other references as appropriate.

azizpunjani:
@notmessenger please assign to me.

PR #741, opened 2015-09-01
Added immediate child selector inplace of children selector

azizpunjani:
closes https://github.com/softlayer/sl-ember-components/issues/727


notmessenger:
All selectors in this file should be changed in addition to the one that you have.

Issue #576, opened 2015-08-11
"DEPRECATION: A property of <dummy@view:-outlet::ember298> was modified inside the didInsertElement hook. You should never change properties on components, services or models during didInsertElement because it causes significant performance degradation.


notmessenger:

juwara0:
I am taking a look at this issue.

juwara0:
This issue can be closed. I have investigated this and am no longer seeing deprecation warnings for this component. My thought is that the changes in https://github.com/azizpunjani/sl-ember-components/blob/c1c184ad36951f83a5d549fe4e0cd0bd82ea0b8e/addon/mixins/sl-tooltip-enabled.js where the dataToggle property and methods setting it were removed may have fixed the issue.

juwara0:
Here is the changeset where dataToggle was removed: https://github.com/softlayer/sl-ember-components/pull/790/files

Issue #554, opened 2015-08-11
"DEPRECATION: A property of <dummy@view:-outlet::ember298> was modified inside the didInsertElement hook. You should never change properties on components, services or models during didInsertElement because it causes significant performance degradation.


notmessenger:

azizpunjani:
@notmessenger please assign to me.

azizpunjani:
@notmessenger please close issue as it has been resolved by https://github.com/softlayer/sl-ember-components/pull/801.

PR #790, opened 2015-09-08
Moved setting of property to init instead of inside didInsertElement

azizpunjani:
Closes https://github.com/softlayer/sl-ember-components/issues/554
Closes https://github.com/softlayer/sl-ember-components/issues/576

### DIFF
--- a/addon/mixins/sl-tooltip-enabled.js
+++ b/addon/mixins/sl-tooltip-enabled.js
@@ -59,6 +59,23 @@ export default Ember.Mixin.create({
     // Observers
 
     /**
+     * Initialize mixin properties
+     *
+     * @function
+     * @returns {undefined}
+     */
+    initialize: Ember.on(
+        'init',
+        function() {
+            if ( this.get( 'popover' ) ) {
+                this.set( 'dataToggle', 'popover' );
+            } else if ( this.get( 'tooltip' ) ) {
+                this.set( 'dataToggle', 'tooltip' );
+            }
+        }
+    ),
+
+    /**
      * Enable the tooltip functionality, based on component's
      * `popover` attribute or component's `title` attribute
      *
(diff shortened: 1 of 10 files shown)


addon/mixins/sl-tooltip-enabled.js
```diff
@@ -96,8 +113,6 @@ export default Ember.Mixin.create({
 
         // First-time rendering
         if ( 'undefined' === Ember.typeOf( originalTitle ) ) {
-            this.set( 'dataToggle', 'popover' );
```
notmessenger: The `dataToggle` value never needs to be changed based on the value(s) of `popover` and `title`?

addon/mixins/sl-tooltip-enabled.js
```diff
@@ -123,8 +138,6 @@ export default Ember.Mixin.create({
 
         // First-time rendering
         if ( 'undefined' === Ember.typeOf( originalTitle ) ) {
-            this.set( 'dataToggle', 'tooltip' );
```
notmessenger: The `dataToggle` value never needs to be changed based on the value(s) of `popover` and `title`?

azizpunjani:
@notmessenger after reading the docs carefully I discovered that we do *not* need the `data-toggle` attribute unless we are initializing the tooltip or popover using a general selector. 
>For performance reasons, the Tooltip and Popover data-apis are opt-in, meaning you must initialize them yourself.
>One way to initialize all tooltips on a page would be to select them by their data-toggle attribute:
```
$(function () {
  $('[data-toggle="tooltip"]').tooltip()
})
```

We initialize the tooltip/popover on the specific element and hence do not need a `data-toggle` attribute for selection purposes. I have removed the `data-toggle` attribute. Let me know if you disagree or find that my conclusions are wrong.

PR #792, opened 2015-09-08
Remove alias to _parentView._controller

### LINKED ISSUE #696: Grid demo is broken
Then change in https://github.com/softlayer/sl-ember-components/issues/556 has caused the demo application at #/demos/sl-grid to no longer work correctly.

jonathandavidson:
This has been accomplished with the following:

1. `rowActionContext` property has been removed from the sl-grid component.
2. Added `dropButtonSelect` action to the sl-grid component which sends an action specified by the second argument to the targetObject.
3. Modified the sl-drop-option component to accept an `actionContext` property which is sent along with the data on click events.
4. Modified the sl-grid template to pass in the action name to the `actionContext` property of each sl-drop-option component on each row.
5. Updated the demo app to show the new usage.
6. Added tests for the new functionality.

Also closes #696


tests/dummy/app/templates/demos/sl-grid.hbs
```diff
@@ -10,13 +10,14 @@
...
     columns=columns
-    content=arrangedContent
+    content=model
     detailComponent="demos/sl-grid-detail"
     detailFooterComponent="demos/sl-grid-detail-footer"
     detailHeaderComponent="demos/sl-grid-detail-header"
     footerPath="demos/sl-grid/footer"
     height=400
     rowActions=rowActions
+    sendLog=(action "sendLog")
```
notmessenger: Does the documentation in the demo need to explain this?  Nothing alludes to how this is all configured.

jonathandavidson: Wouldn't hurt.

Issue #768, opened 2015-09-03
Create Unit test suite


notmessenger:
Test:

* default values
* `radioType()`
* properties observed by computed property

erangeles:
@notmessenger please assign this to me.

PR #799, opened 2015-09-09
 Create Unit test suite #768

erangeles:


tests/unit/components/sl-radio-test.js
```diff
@@ -0,0 +1,48 @@
+import{ moduleForComponent, test } from 'ember-qunit';
+
+moduleForComponent( 'sl-radio', 'Unit | Component | sl radio', {
+    unit: true
+});
+
+test( 'Correct default property values', function( assert ) {
```
notmessenger: Should also assert value of `tagName` property

tests/unit/components/sl-radio-test.js
```diff
@@ -0,0 +1,48 @@
...
+    );
+
+    assert.strictEqual(
+        component.get( 'value' ),
+        null,
+        'Default property "value" is null'
+    );
+});
+
+test( 'Inline property sets relevant class', function( assert ) {
```
notmessenger: * This test should be testing both conditions:  "radio-inline" when inline, "radio" default
* The `_dependentKeys` property should be used to confirm the correct properties are being observed.  A demo of this property's use (not in a test though approach is the same) can be viewed at http://jsbin.com/zituta/edit?html,js,output
* This computed property should be tested via calling it, such as via `this.get( 'radioType` )`.  Integration tests are about the rendering of the component and whether the template is configured as expected, unit tests are about the output of the function.

erangeles:
@notmessenger having issues obtaining `_dependentKeys` as it's an "internal" variable. I only get the value of the property when I use the `this.get( 'radioType' )`. everything else has been addressed.

notmessenger:
@erangeles

Does this work?

```
const component = this.subject();

component.radioType._dependentKeys
```

erangeles:
@notmessenger all comments have been address; PR is ready to be reviewed again.

PR #801, opened 2015-09-10
Deprecation warning fix

azizpunjani:
@notmessenger after reading the [docs](http://getbootstrap.com/javascript/#tooltips) carefully I discovered that we do *not* need the `data-toggle` attribute unless we are initializing the tooltip or popover using a general attribute selector. 
>For performance reasons, the Tooltip and Popover data-apis are opt-in, meaning you must initialize them yourself.
>One way to initialize all tooltips on a page would be to select them by their data-toggle attribute:
```
$(function () {
  $('[data-toggle="tooltip"]').tooltip()
})
```

We initialize the tooltip/popover for each specific element.
`this.$().tooltip({....});`

Hence do not need a `data-toggle` attribute for selection purposes. I have removed the `data-toggle` attribute. Let me know if you disagree or find that my conclusions are wrong.

### DIFF
--- a/addon/mixins/sl-tooltip-enabled.js
+++ b/addon/mixins/sl-tooltip-enabled.js
@@ -59,6 +59,23 @@ export default Ember.Mixin.create({
     // Observers
 
     /**
+     * Initialize mixin properties
+     *
+     * @function
+     * @returns {undefined}
+     */
+    initialize: Ember.on(
+        'init',
+        function() {
+            if ( this.get( 'popover' ) ) {
+                this.set( 'dataToggle', 'popover' );
+            } else if ( this.get( 'tooltip' ) ) {
+                this.set( 'dataToggle', 'tooltip' );
+            }
+        }
+    ),
+
+    /**
      * Enable the tooltip functionality, based on component's
      * `popover` attribute or component's `title` attribute
      *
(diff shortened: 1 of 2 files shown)


notmessenger:
http://getbootstrap.com/javascript/#markup-1

> The required markup for a tooltip is only a data attribute and title on the HTML element you wish to have a tooltip. The generated markup of a tooltip is rather simple, though it does require a position (by default, set to top by the plugin).

```
<!-- HTML to write -->
<a href="#" data-toggle="tooltip" title="Some tooltip text!">Hover over me</a>

<!-- Generated markup by the plugin -->
<div class="tooltip top" role="tooltip">
  <div class="tooltip-arrow"></div>
  <div class="tooltip-inner">
    Some tooltip text!
  </div>
</div>
```

Doesn't this indicate that the `data-toggle` attribute is required, regardless of how we are initializing it?

azizpunjani:
@notmessenger I don't believe so. The documentation is misleading. If you look at the options, you will notice there isn't a `data-toggle` option. http://getbootstrap.com/javascript/#tooltips-options

azizpunjani:
@notmessenger will do a little more research on this.

notmessenger:
@azizpunjani Based on your reply and what I read here http://stackoverflow.com/questions/17255367/bootstrap-tooltip-data-toggle I am in agreement.

PR #805, opened 2015-09-11
Closes softlayer/sl-ember-components#550

juwara0:
Fix deprecation message in sl-date-range-picker by incorporating the sl-component-input-id mixin, updated component template, docs and tests.

PR #806, opened 2015-09-11
Closes softlayer/sl-ember-components#553

juwara0:
Fix deprecation message in sl-date-picker by incorporating the sl-component-input-id mixin, updated component template, docs and tests.

### DIFF
--- a/tests/unit/components/sl-date-picker-test.js
+++ b/tests/unit/components/sl-date-picker-test.js
@@ -1,9 +1,17 @@
 import Ember from 'ember';
 import { moduleForComponent, test } from 'ember-qunit';
 import sinon from 'sinon';
+import ComponentInputId from 'sl-ember-components/mixins/sl-component-input-id';
 
 moduleForComponent( 'sl-date-picker', 'Unit | Component | sl date picker', {
     unit: true
+});
+
+test( 'Expected Mixins are present', function( assert ) {
+    assert.ok(
+        ComponentInputId.detect( this.subject() ),
+        'ComponentInputId Mixin is present'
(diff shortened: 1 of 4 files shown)


tests/unit/components/sl-date-picker-test.js
```diff
@@ -1,11 +1,19 @@
...
+import ComponentInputId from 'sl-ember-components/mixins/sl-component-input-id';
 
 moduleForComponent( 'sl-date-picker', 'Unit | Component | sl date picker', {
     unit: true
 });
 
+test( 'Expected Mixins are present', function( assert ) {
+    assert.ok(
+        ComponentInputId.detect( this.subject() ),
+        'ComponentInputId Mixin is present'
```
notmessenger: If going to reference the mixin name, would use the actual Mixin name, not the import alias.

notmessenger:
Failing test:

```
not ok 251 PhantomJS 1.9 - Unit | Component | sl date range picker: label is accepted as a parameter
    ---
        actual: >
            
        expected: >
            null
        message: >
            label element has the correct for property
        Log: |
    ...
```

juwara0:
In response to: 
Failing test:

not ok 251 PhantomJS 1.9 - Unit | Component | sl date range picker: label is accepted as a parameter
    ---
        actual: >

        expected: >
            null
        message: >
            label element has the correct for property
        Log: |
    ...

This expected because I made a change to sl-date-picker and sl-date-range-picker is composed of two date-pickers. The update to address this is in: https://github.com/softlayer/sl-ember-components/pull/805

Issue #802, opened 2015-09-10
Refactor "spellcheck" property


notmessenger:
* is set to an invalid default value
* its list of acceptable values in the DocBlock do not align with supported values in the `enum`

erangeles:
@notmessenger taking this as it's in tandem with #605.

erangeles:
@notmessenger, this is no longer an issue, has been addressed already. https://github.com/softlayer/sl-ember-components/blob/EmberCli-1.13.8/addon/components/sl-textarea.js#L98 https://github.com/softlayer/sl-ember-components/blob/EmberCli-1.13.8/tests/dummy/app/templates/demos/sl-textarea.hbs#L127 please move this into the done section of waffle.

PR #705, opened 2015-08-26
 Create integration test suite and migrate appropriate tests from unit test suite #605

erangeles:
@juwara0 and I will be pairing to resolve an issue with selectionEnd/Start/Direction properties in textArea as they might have been removed from ember textArea. Just FYI, the rest is ready to be reviewed.

[Review on Reviewable](https://reviewable.io/reviews/softlayer/sl-ember-components/705)

PR #823, opened 2015-09-11
 Refactor "spellcheck" property #802 

erangeles:
In this PR it was decided to remove the Spellcheck enum from sl-textarea component.

The enum that represented the Spellcheck property was originally defaulted to the value "default" which is a wrong/invalid value. Removing that would leave the enum redefining a boolean so ultimately we removed that enum all together as it served no purpose. Essentially we set the defaulted value of the Spellcheck property to false because a false and an absence of an attribute are equivalent according to w3. source below.

Boolean attributes are explained here: http://www.w3.org/TR/html4/intro/sgmltut.html#h-3.3.4.2

Some attributes play the role of boolean variables (e.g., the selected attribute for the OPTION element). Their appearance in the start tag of an element implies that the value of the attribute is "true". Their absence implies a value of "false".

### DIFF
--- a/addon/components/sl-textarea.js
+++ b/addon/components/sl-textarea.js
@@ -94,7 +94,7 @@ export default Ember.Component.extend( InputBased, TooltipEnabled, ComponentInpu
      *
      * Accepted values "false" (default) or "true".
      *
-     * @type {Spellcheck}
+     * @type {Boolean}
      */
     spellcheck: false,
 
--- a/addon/components/sl-textarea.js
+++ b/addon/components/sl-textarea.js
@@ -92,8 +92,6 @@ export default Ember.Component.extend( InputBased, TooltipEnabled, ComponentInpu
     /**
      * The `spellcheck` HTML attribute value
      *
-     * Accepted values "false" (default) or "true".
-     *
      * @type {Boolean}
      */
     spellcheck: false,
@@ -119,5 +117,4 @@ export default Ember.Component.extend( InputBased, TooltipEnabled, ComponentInpu
 
     // -------------------------------------------------------------------------
     // Methods
-
 });
--- a/addon/components/sl-textarea.js
+++ b/addon/components/sl-textarea.js
@@ -17,18 +17,6 @@ export const Direction = Object.freeze({
 });
 
 /**
- * Valid values for `spellcheck` property
- *
- * @memberof module:components/sl-textarea
- * @enum {Boolean|String}
- */
-export const Spellcheck = {
-    DEFAULT: 'default',
-    FALSE: false,
-    TRUE: true
-};
-
-/**
  * Valid values for `wrap` property
  *
  * @memberof module:components/sl-textarea
@@ -104,11 +92,11 @@ export default Ember.Component.extend( InputBased, TooltipEnabled, ComponentInpu
     /**
      * The `spellcheck` HTML attribute value
      *
-     * Accepted values are true, false, "default" (default), "true", or "false".
+     * Accepted values "false" (default) or "true".
      *
      * @type {Spellcheck}
      */
-    spellcheck: Spellcheck.Direction,
+    spellcheck: false,
 
     /**
      * The bound value of the textarea
--- a/tests/dummy/app/templates/demos/sl-textarea.hbs
+++ b/tests/dummy/app/templates/demos/sl-textarea.hbs
@@ -123,8 +123,8 @@
         The <code>selectionStart</code> HTML attribute value, passed directly to the underlying &lt;textarea&gt;.
     {{/property-text}}
 
-    {{#property-text name="spellcheck" type="Boolean/String" default="\"default\""}}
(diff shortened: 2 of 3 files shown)


notmessenger:
> In this PR it was decided to remove the Spellcheck property.

We're not removing the property are we? Aren't we only removing the `enum` used in association with it?

notmessenger:
Where are the associated test changes?

addon/components/sl-textarea.js
```diff
@@ -104,11 +92,11 @@ export default Ember.Component.extend( InputBased, TooltipEnabled, ComponentInpu
...
      * The `spellcheck` HTML attribute value
      *
-     * Accepted values are true, false, "default" (default), "true", or "false".
+     * Accepted values "false" (default) or "true".
      *
-     * @type {Spellcheck}
+     * @type {Boolean}
      */
-    spellcheck: Spellcheck.Direction,
+    spellcheck: false,
```
notmessenger: Have you thoroughly tested this with bound values in an app scenario?  Boolean `false` is different than a string `"false` (and likewise for `true`/ `"true"`).  Are we sure we're getting the desired behavior out of this configuraiton?

addon/components/sl-textarea.js
```diff
@@ -104,11 +92,11 @@ export default Ember.Component.extend( InputBased, TooltipEnabled, ComponentInpu
     /**
      * The `spellcheck` HTML attribute value
      *
-     * Accepted values are true, false, "default" (default), "true", or "false".
+     * Accepted values "false" (default) or "true".
```
notmessenger: If `@type` is to remain a Boolean, this line is unnecessary.

tests/dummy/app/templates/demos/sl-textarea.hbs
```diff
@@ -123,8 +123,8 @@
         The <code>selectionStart</code> HTML attribute value, passed directly to the underlying &lt;textarea&gt;.
     {{/property-text}}
 
-    {{#property-text name="spellcheck" type="Boolean/String" default="\"default\""}}
-        The <code>spellcheck</code> HTML attribute value, passed directly to the underlying &lt;textarea&gt;. Can be true, false, or "default".
+    {{#property-text name="spellcheck" type="Boolean" default="\"false\""}}
```
notmessenger: If this is a Boolean type why is its default value a string?

tests/dummy/app/templates/demos/sl-textarea.hbs
```diff
@@ -123,8 +123,8 @@
...
+        The <code>spellcheck</code> HTML attribute value, passed directly to the underlying &lt;textarea&gt;. Can be true, false.
```
notmessenger: If the type is a Boolean, the text of "Can be true, false." seems unnecessary.

addon/components/sl-textarea.js
erangeles: The type is specified to be Boolean so there is no need to handle string values, its not our fault that users pass in a wrong type, because our documentation states a Boolean value. we should not be testing for every single contingency. I tested a computed value called spellCheckString that converted string false to boolean false and vice versa like how the input based mixin does with the disabled property https://github.com/softlayer/sl-ember-components/blob/EmberCli-1.13.8/addon/mixins/sl-input-based.js#L98 , however that is not correct in this case.

erangeles:
@notmessenger testing was added to issue https://github.com/softlayer/sl-ember-components/pull/705/files but will be removed from there and added here instead.

erangeles:
@notmessenger all comments addressed, it can be reviewed again.

tests/integration/components/sl-textarea-test.js
```diff
@@ -17,3 +17,57 @@ test( 'for attribute value on label matches id of textarea', function( assert )
...
+test( '"spellcheck" property is supported', function( assert ) {
+    this.render( hbs`
+        {{#sl-textarea spellcheck=true}}
+        {{/sl-textarea}}
+    ` );
+
+    const wrapper = this.$( '>:first-child' );
+
+    assert.strictEqual(
+        wrapper.find( 'textarea' ).attr( 'spellcheck' ),
```
notmessenger: Replace `wrapper` with `this.$( '>:first-child' )` to be consistent with the rest of the codebase.

tests/integration/components/sl-textarea-test.js
```diff
@@ -17,3 +17,57 @@ test( 'for attribute value on label matches id of textarea', function( assert )
...
+test( '"spellcheck" property defaults correctly', function( assert ) {
+    this.render( hbs`
+        {{#sl-textarea}}
+        {{/sl-textarea}}
+    ` );
+
+    const wrapper = this.$( '>:first-child' );
+
+    assert.strictEqual(
+        wrapper.find( 'textarea' ).attr( 'spellcheck' ),
```
notmessenger: Replace `wrapper` with `this.$( '>:first-child' )` to be consistent with the rest of the codebase.

Issue #594, opened 2015-08-11
Create integration test suite and migrate appropriate tests from unit test suite


notmessenger:

jonathandavidson:
Picking this up

PR #938, opened 2015-10-15
Refactor tests for sl-pagination and add integration test suite #594

jonathandavidson:
[Review on Reviewable](https://reviewable.io/reviews/softlayer/sl-ember-components/938)

### DIFF
--- a/tests/integration/components/sl-pagination-test.js
+++ b/tests/integration/components/sl-pagination-test.js
@@ -1,0 +1,205 @@
+import { moduleForComponent, test } from 'ember-qunit';
+import hbs from 'htmlbars-inline-precompile';
+import sinon from 'sinon';
+
+moduleForComponent( 'sl-pagination', 'Integration | Component | sl pagination', {
+    integration: true
+});
+
+test( 'Default classes are applied', function( assert ) {
+    this.render( hbs`
+        {{sl-pagination totalPages=1}}
+    ` );
+
+    assert.ok(
+        this.$( '>:first-child' ).hasClass( 'pagination' ),
+        'Default rendered component has class "pagination"'
+    );
+
+    assert.ok(
+        this.$( '>:first-child' ).hasClass( 'sl-pagination' ),
+        'Default rendered component has class "sl-pagination"'
+    );
+});
+
+test( 'The totalPages property is bound to the total pages display', function( assert ) {
+    this.set( 'totalPages', 1 );
+
+    this.render( hbs`
+        {{sl-pagination totalPages=totalPages}}
+    ` );
+
+    assert.strictEqual(
+        this.$( '>:first-child' ).find( 'li:nth-child(2) a' ).text().trim(),
+        '1 / 1',
+        'totalPages is initialized to 1'
+    );
+
+    this.set( 'totalPages', 2 );
+
+    assert.strictEqual(
+        this.$( '>:first-child' ).find( 'li:nth-child(2) a' ).text().trim(),
+        '1 / 2',
+        'totalPages is now set to 2'
+    );
+});
+
+test( 'The currentPage property is bound to the current page display', function( assert ) {
+    this.set( 'currentPage', 1 );
+
+    this.render( hbs`
+        {{sl-pagination totalPages=2 currentPage=currentPage}}
+    ` );
+
+    assert.strictEqual(
+        this.$( '>:first-child' ).find( 'li:nth-child(2) a' ).text().trim(),
+        '1 / 2',
+        'currentPage is initialized to 1'
+    );
+
+    this.set( 'currentPage', 2 );
+
+    assert.strictEqual(
+        this.$( '>:first-child' ).find( 'li:nth-child(2) a' ).text().trim(),
+        '2 / 2',
+        'currentPage is now set to 2'
+    );
+});
+
+test( 'When totalPages is 1 the previous and next buttons are disabled', function( assert ) {
+    this.render( hbs`
+        {{sl-pagination totalPages=1}}
+    ` );
+
+    assert.ok(
+        this.$( '>:first-child' ).find( 'li:first-child' ).hasClass( 'disabled' ),
+        'The previous button has the "disabled" class'
+    );
+
+    assert.ok(
+        this.$( '>:first-child' ).find( 'li:nth-child(3)' ).hasClass( 'disabled' ),
+        'The next button has the "disabled" class'
+    );
+});
+
+test( 'The previous button is disabled when on the first page', function( assert ) {
+    const testActionHandler = sinon.spy();
+
+    this.on( 'testAction', testActionHandler );
+
+    this.render( hbs`
+        {{sl-pagination totalPages=2 changePage="testAction"}}
+    ` );
+
+    assert.ok(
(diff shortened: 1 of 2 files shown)


notmessenger:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

<sup>**[tests/unit/components/sl-pagination-test.js, line 62 \[r1\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/938#-K1AhsMcPQuY0-la3w8E)** ([raw file](https://github.com/softlayer/sl-ember-components/blob/c04dc520de47d6c48d0959fc28642a2006022eca/tests/unit/components/sl-pagination-test.js#L62)):</sup>
This test case should only assert that `changePageBy()` is called with the value of `1`


---

<sup>**[tests/unit/components/sl-pagination-test.js, line 76 \[r1\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/938#-K1Ai-etoeo9fCJS6eh5)** ([raw file](https://github.com/softlayer/sl-ember-components/blob/c04dc520de47d6c48d0959fc28642a2006022eca/tests/unit/components/sl-pagination-test.js#L76)):</sup>
This test case should only assert that `changePageBy()` is called with the value of `-1`


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/938)

notmessenger:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/938)

jonathandavidson:
Review status: all files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

<sup>**[tests/unit/components/sl-pagination-test.js, line 62 \[r1\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/938#-K1AhsMcPQuY0-la3w8E)** ([raw file](https://github.com/softlayer/sl-ember-components/blob/c04dc520de47d6c48d0959fc28642a2006022eca/tests/unit/components/sl-pagination-test.js#L62)):</sup>
It feels wrong to be stubbing a method of this same class. This post seems to confirm my reasoning: http://programmers.stackexchange.com/questions/107361/is-it-okay-to-fake-part-of-the-class-under-test. By that standard this test would be correct and the tests directly testing changePageBy() should be modified.  Either way I need to make some changes but I'd rather do this in a way that causes us the least problems down the road. Do you have a strong opinion on this?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/938)

jonathandavidson:
Review status: all files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

<sup>**[tests/unit/components/sl-pagination-test.js, line 76 \[r1\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/938#-K1Ai-etoeo9fCJS6eh5)** ([raw file](https://github.com/softlayer/sl-ember-components/blob/c04dc520de47d6c48d0959fc28642a2006022eca/tests/unit/components/sl-pagination-test.js#L76)):</sup>
See my other comment regarding stubbing class methods.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/938)

notmessenger:
<sup>**[tests/unit/components/sl-pagination-test.js, line 62 \[r1\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/938#-K1AhsMcPQuY0-la3w8E)** ([raw file](https://github.com/softlayer/sl-ember-components/blob/c04dc520de47d6c48d0959fc28642a2006022eca/tests/unit/components/sl-pagination-test.js#L62)):</sup>
I am not certain how to apply what you linked to the situation at hand.  `changePageBy()` is already thoroughly tested, which it should be.  The only logic in the `nextPage` and `previousPage` actions is calling `changePageBy()` with different values.  Since the other tests have taken care of `changePageBy()`'s functionality none of it, or its side-effects, need to be tested.  We only care we are passing the correct value to it within the respective actions.  Other code is taking care of the rest of this.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/938)

notmessenger:
<sup>**[tests/unit/components/sl-pagination-test.js, line 76 \[r1\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/938#-K1Ai-etoeo9fCJS6eh5)** ([raw file](https://github.com/softlayer/sl-ember-components/blob/c04dc520de47d6c48d0959fc28642a2006022eca/tests/unit/components/sl-pagination-test.js#L76)):</sup>
See my other comment


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/938)

jonathandavidson:
Review status: all files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

<sup>**[tests/unit/components/sl-pagination-test.js, line 62 \[r1\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/938#-K1AhsMcPQuY0-la3w8E)** ([raw file](https://github.com/softlayer/sl-ember-components/blob/c04dc520de47d6c48d0959fc28642a2006022eca/tests/unit/components/sl-pagination-test.js#L62)):</sup>
Maybe that wasn't the clearest example but what I'm getting at is that we are setting ourselves up for headaches if we are testing "private" methods directly and/or stubbing them in other tests on the same class. I realize that `changePageBy()` is technically not private, but it isn't a public requirement and therefore is an implementation detail. Say the component was refactored to break up `changePageBy()` into two separate methods, `incrementPage()` and `decrementPage()`. Then we would have to rewrite our entire unit test suite for this component. If on the other hand we are using a black box testing approach, after the refactor all of the tests will still pass, so that we can verify that we did the refactor correctly. This is after all the whole reason we write tests in the first place.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/938)

notmessenger:
Review status: all files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

<sup>**[tests/unit/components/sl-pagination-test.js, line 62 \[r1\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/938#-K1AhsMcPQuY0-la3w8E)** ([raw file](https://github.com/softlayer/sl-ember-components/blob/c04dc520de47d6c48d0959fc28642a2006022eca/tests/unit/components/sl-pagination-test.js#L62)):</sup>
Yeah, I see what you're saying.  Specifically in this example it makes perfect sense to do what you describe.  In other actions in other components there is often a lot more going on.  Even if this logic is broken into individual method calls it is still burdensome to test the end result of all of it rather than just that expected things are called, especially when there are logic branches around the calls.  I know that you are not saying to make this a carte blanche pattern we follow, but I would say that the approach to use should be evaluated on a case-by-case basis and the best one chosen.

For this scenario, as presented, I am fine with your approach and will merge this PR.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/938)

Issue #746, opened 2015-09-02
Is there better syntax to accomplish this?


notmessenger:
I could certainly be wrong on this matter but I seem to recall there might be a better way to get a reference to the registry than by using `this.get( 'container._registry' );` that we currently do in `setupTemplates()` in addon/components/sl-grid.js

Also look at the approach used with `this.get( '_parentView.renderedName' );` in the same method.

PR #1017, opened 2015-10-26
Closes softlayer/sl-ember-components#746

SpikedKira:
It seems like we don't need any of this code any more.  Likely new versions of Ember added support for partials that was previously lacking.

[Review on Reviewable](https://reviewable.io/reviews/softlayer/sl-ember-components/1017)

### DIFF
--- a/addon/components/sl-grid.js
+++ b/addon/components/sl-grid.js
@@ -441,32 +441,6 @@ export default Ember.Component.extend({
     ),
 
     /**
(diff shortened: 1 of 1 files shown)


notmessenger:
@joshforisha Can you please review this?

joshforisha:
Looks good to me. The partials that the grid was previously using were changed to sub-components by @jonathandavidson I believe, so this setup should be unnecessary at this point.

Issue #852, opened 2015-09-22
Is our implementation requiring CSP to include `unsafe-inline`?


notmessenger:
If so, consider using https://github.com/BryanCrotaz/ember-cli-csp-style, which should also be referred to for context to this issue.  Once context is obtained, test our implementation to see if this problem exists.

juwara0:
@notmessenger: I am working on this.

PR #1152, opened 2015-11-11
Implemented fix for CSP warning

juwara0:
Closes softlayer/sl-ember-components#852

[Review on Reviewable](https://reviewable.io/reviews/softlayer/sl-ember-components/1152)

### DIFF
--- a/addon/components/sl-progress-bar-indicator.js
+++ b/addon/components/sl-progress-bar-indicator.js
@@ -1,0 +1,82 @@
+import Ember from 'ember';
+import layout from '../templates/components/sl-progress-bar-indicator';
+import CspStyle from 'ember-cli-csp-style/mixins/csp-style';
+
+/**
+ * @module
+ * @augments ember/Component
+ * @augments ember-cli-csp-style/mixins/csp-style
+*/
+export default Ember.Component.extend( CspStyle, {
+
+    // -------------------------------------------------------------------------
+    // Dependencies
+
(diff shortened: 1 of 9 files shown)


juwara0:
@notmessenger: This is ready for review

azizpunjani:
Personally I think we could have avoided the CSP mixin and creating another component.  

Here's how I think we could have avoided the mixin and the extra component https://gist.github.com/azizpunjani/224920f343747befaaaf. This is just from my initial investigation and 10minutes of tinkering with CSP warnings. There might be a valid use case for using the CSP mixin. If this is the way that was agreed upon I am OK with it.


---

Review status: 0 of 9 files reviewed at latest revision, all discussions resolved.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1152)

azizpunjani:
Reviewed 8 of 9 files at r1, 1 of 1 files at r2.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1152)

notmessenger:
@juwara0 This does seem like a simpler approach but I may be missing something


---

Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1152)

Issue #1122, opened 2015-11-10
Style component using Bootstrap default structure and styles


notmessenger:
Replace use of font awesome classes with "generic" ones and then use CSS declarations to apply Glyphicon styling.

PR #1209, opened 2015-11-20
Sl calendar icons

azizpunjani:
Was initially able to use bootstrap glyphicons for next and previous. The problem we run into is that sl-datepicker ends up looking different because it does not use glyphicons. I tried to override sl-datepicker to use glyphicons but it ended up being a lot of CSS. 

I opted to make sl-calendar consistent with sl-datepicker by simply using the same htmlentities that sl-datepicker uses, instead of applying over 30 lines of CSS to override the sl-datepicker styles.

Closes https://github.com/softlayer/sl-ember-components/issues/1122

[Review on Reviewable](https://reviewable.io/reviews/softlayer/sl-ember-components/1209)

### DIFF
--- a/addon/components/sl-calendar-day.js
+++ b/addon/components/sl-calendar-day.js
@@ -13,11 +13,17 @@ export default Ember.Component.extend({
     // -------------------------------------------------------------------------
     // Attributes
 
+    ariaRole: 'button',
+
+    attributeBindings: [
+        'tabindex'
+    ],
+
     /** @type {String[]} */
     classNameBindings: [
         'active',
         'new',
-        'old'
+        'old',
     ],
 
     /** @type {String[]} */
@@ -27,6 +33,8 @@ export default Ember.Component.extend({
 
     /** @type {Object} */
     layout,
+
+    tabindex: 0,
 
     /** @type {String} */
     tagName: 'td',
--- a/addon/components/sl-calendar-day.js
+++ b/addon/components/sl-calendar-day.js
@@ -17,7 +17,7 @@ export default Ember.Component.extend({
     classNameBindings: [
         'active',
         'new',
-        'old',
+        'old'
     ],
 
     /** @type {String[]} */
--- a/addon/components/sl-calendar-day.js
+++ b/addon/components/sl-calendar-day.js
@@ -13,12 +13,6 @@ export default Ember.Component.extend({
     // -------------------------------------------------------------------------
     // Attributes
 
-    ariaRole: 'button',
-
-    attributeBindings: [
-        'tabindex'
-    ],
-
     /** @type {String[]} */
     classNameBindings: [
         'active',
@@ -33,8 +27,6 @@ export default Ember.Component.extend({
 
     /** @type {Object} */
     layout,
-
-    tabindex: 0,
 
     /** @type {String} */
     tagName: 'td',
--- a/addon/components/sl-calendar.js
+++ b/addon/components/sl-calendar.js
@@ -20,7 +20,8 @@ export default Ember.Component.extend({
 
     /** @type {String[]} */
     classNames: [
-        'sl-calendar'
+        'sl-ember-components',
+        'calendar'
     ],
 
     /** @type {Object} */
--- a/addon/templates/components/sl-calendar.hbs
+++ b/addon/templates/components/sl-calendar.hbs
@@ -1,17 +1,17 @@
 <div class="datepicker datepicker-inline">
     {{#if viewingDays}}
-        <div class="datepicker-days" style="display: block;">
+        <div class="datepicker-days">
             <table class="table-condensed">
                 <thead>
                     <tr>
                         <th {{action "changeMonth" -1}} class="sl-icon-prev">
-                            <span class="sr-only">previous</span>
+                            previous month
                         </th>
-                        <th {{action "setView" "months"}} colspan="5" class="datepicker-switch">
+                        <th {{action "setView" "months"}} class="datepicker-switch" colspan="5">
                             {{currentMonthString}} {{currentYear}}
                         </th>
                         <th {{action "changeMonth" 1}} class="sl-icon-next">
-                            <span class="sr-only">next</span>
+                            next month
                         </th>
                     </tr>
                     <tr>
@@ -41,18 +41,18 @@
     {{/if}}
 
     {{#if viewingMonths}}
-        <div class="datepicker-months" style="display: block;">
+        <div class="datepicker-months">
             <table class="table-condensed">
                 <thead>
                     <tr>
-                        <th {{action "changeYear" -1}} class="sl-icon-prev" style="visibility: visible;">
-                            <span class="sr-only">previous</span>
+                        <th {{action "changeYear" -1}} class="sl-icon-prev">
+                            previous year
                         </th>
                         <th {{action "setView" "years"}} class="datepicker-switch" colspan="5">
                             {{currentYear}}
                         </th>
-                        <th {{action "changeYear" 1}} class="sl-icon-next" style="visibility: visible;">
-                            <span class="sr-only">next</span>
+                        <th {{action "changeYear" 1}} class="sl-icon-next">
+                            next year
                         </th>
                     </tr>
                 </thead>
@@ -75,18 +75,18 @@
     {{/if}}
 
     {{#if viewingYears}}
-        <div class="datepicker-years" style="display: block;">
+        <div class="datepicker-years">
             <table class="table-condensed">
                 <thead>
                     <tr>
-                        <th {{action "changeDecade" -1}} class="sl-icon-prev" style="visibility: visible;">
-                            <span class="sr-only">previous</span>
+                        <th {{action "changeDecade" -1}} class="sl-icon-prev">
+                            previous decade
                         </th>
                         <th class="datepicker-switch" colspan="5">
                             {{decadeStart}}-{{decadeEnd}}
                         </th>
-                        <th {{action "changeDecade" 1}} class="sl-icon-next" style="visibility: visible;">
-                            <span class="sr-only">next</span>
+                        <th {{action "changeDecade" 1}} class="sl-icon-next">
+                            next decade
                         </th>
                     </tr>
                 </thead>
--- a/addon/templates/components/sl-calendar.hbs
+++ b/addon/templates/components/sl-calendar.hbs
@@ -5,7 +5,7 @@
                 <thead>
                     <tr>
                         <th {{action "changeMonth" -1}} class="prev">
-                            <span class="fa fa-angle-left"></span>
+                            &#171;
                         </th>
                         <th
                             {{action "setView" "months"}}
@@ -15,7 +15,7 @@
                             {{currentMonthString}} {{currentYear}}
                         </th>
                         <th {{action "changeMonth" 1}} class="next">
-                            <span class="fa fa-angle-right"></span>
+                            &#187;
                         </th>
                     </tr>
                     <tr>
@@ -54,7 +54,7 @@
                             class="prev"
                             style="visibility: visible;"
                         >
-                            <span class="fa fa-angle-left"></span>
+                        &#171;
                         </th>
                         <th
                             {{action "setView" "years"}}
@@ -68,7 +68,7 @@
                             class="next"
                             style="visibility: visible;"
                         >
-                            <span class="fa fa-angle-right"></span>
+                            &#187;
                         </th>
                     </tr>
                 </thead>
@@ -97,7 +97,7 @@
                     <tr>
                         <th
                             {{action "changeDecade" -1}} class="prev" style="visibility: visible;">
-                            <span class="fa fa-angle-left"></span>
+                            &#171;
                         </th>
                         <th
                             class="datepicker-switch"
@@ -110,7 +110,7 @@
                             class="next"
                             style="visibility: visible;"
                         >
-                            <span class="fa fa-angle-right"></span>
+                        &#187;
                         </th>
                     </tr>
                 </thead>
--- a/addon/templates/components/sl-calendar.hbs
+++ b/addon/templates/components/sl-calendar.hbs
@@ -110,7 +110,7 @@
                             class="next"
                             style="visibility: visible;"
                         >
-                        &#187;
+                            &#187;
                         </th>
                     </tr>
                 </thead>
--- a/addon/templates/components/sl-calendar.hbs
+++ b/addon/templates/components/sl-calendar.hbs
@@ -4,18 +4,14 @@
             <table class="table-condensed">
                 <thead>
                     <tr>
-                        <th {{action "changeMonth" -1}} class="prev">
-                            &#171;
+                        <th {{action "changeMonth" -1}} class="sl-icon-prev">
+                            <span class="sr-only">previous</span>
                         </th>
-                        <th
-                            {{action "setView" "months"}}
-                            colspan="5"
-                            class="datepicker-switch"
-                        >
+                        <th {{action "setView" "months"}} colspan="5" class="datepicker-switch">
                             {{currentMonthString}} {{currentYear}}
                         </th>
-                        <th {{action "changeMonth" 1}} class="next">
-                            &#187;
+                        <th {{action "changeMonth" 1}} class="sl-icon-next">
+                            <span class="sr-only">next</span>
                         </th>
                     </tr>
                     <tr>
@@ -49,26 +45,14 @@
             <table class="table-condensed">
                 <thead>
                     <tr>
-                        <th
-                            {{action "changeYear" -1}}
-                            class="prev"
-                            style="visibility: visible;"
-                        >
-                        &#171;
+                        <th {{action "changeYear" -1}} class="sl-icon-prev" style="visibility: visible;">
+                            <span class="sr-only">previous</span>
                         </th>
-                        <th
-                            {{action "setView" "years"}}
-                            class="datepicker-switch"
-                            colspan="5"
-                        >
+                        <th {{action "setView" "years"}} class="datepicker-switch" colspan="5">
                             {{currentYear}}
                         </th>
-                        <th
-                            {{action "changeYear" 1}}
-                            class="next"
-                            style="visibility: visible;"
-                        >
-                            &#187;
+                        <th {{action "changeYear" 1}} class="sl-icon-next" style="visibility: visible;">
+                            <span class="sr-only">next</span>
                         </th>
                     </tr>
                 </thead>
@@ -95,22 +79,14 @@
             <table class="table-condensed">
                 <thead>
                     <tr>
-                        <th
-                            {{action "changeDecade" -1}} class="prev" style="visibility: visible;">
-                            &#171;
+                        <th {{action "changeDecade" -1}} class="sl-icon-prev" style="visibility: visible;">
+                            <span class="sr-only">previous</span>
                         </th>
-                        <th
-                            class="datepicker-switch"
-                            colspan="5"
-                        >
+                        <th class="datepicker-switch" colspan="5">
                             {{decadeStart}}-{{decadeEnd}}
                         </th>
-                        <th
-                            {{action "changeDecade" 1}}
-                            class="next"
-                            style="visibility: visible;"
-                        >
-                            &#187;
+                        <th {{action "changeDecade" 1}} class="sl-icon-next" style="visibility: visible;">
+                            <span class="sr-only">next</span>
                         </th>
                     </tr>
                 </thead>
--- a/addon/templates/components/sl-calendar.hbs
+++ b/addon/templates/components/sl-calendar.hbs
@@ -5,13 +5,13 @@
                 <thead>
                     <tr>
                         <th {{action "changeMonth" -1}} class="sl-icon-prev">
-                            previous month
+                            <span class="sr-only">previous month</span>
                         </th>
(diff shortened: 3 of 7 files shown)


notmessenger:
For each of the spans that were replaced with HTML entities, replace instead with a span with a class of "sl-icon-prev" (and appropriate class name based on specific context for each).  Then add this (pseudo) CSS declaration:

```
.calendar .sl-icon-prev:prev { content: "&#171"; font-family: serif; }
```

@SpikedKira and myself know the reasoning behind this and will gladly share it "in person" if desired.  Too much to write here.

---

Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-:-K4ZCLEarH89UQ26yrKu:848304177)*

SpikedKira:
.sl-ember-components.calendar .sl-icon-prev:before { content: "&#171;"; font-family: serif; }

that would be the more preferred :) I typed it lazily in chat with @notmessenger

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-:-K4ZM9VOi5_J8PZEOabU:1800216295)*

notmessenger:
There should be assertions or test case added to the integration test suite for sr-only content.

---

Reviewed 5 of 5 files at r2, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[addon/components/sl-calendar-day.js, line 16 \[r2\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-K4hL8Cb3408SOvkw-dP:-K4hL8Cb3408SOvkw-dQ:751400783) ([raw file](https://github.com/softlayer/sl-ember-components/blob/88bd9102a242a3b5b93cf597e08a3b9acbd2982c/addon/components/sl-calendar-day.js#L16)):*
Belongs in the "Properties" comment block and requires a DocBlock

---

*[addon/components/sl-calendar-day.js, line 18 \[r2\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-K4hLDpMqcie3T1H73ME:-K4hLDpMqcie3T1H73MF:-718837343) ([raw file](https://github.com/softlayer/sl-ember-components/blob/88bd9102a242a3b5b93cf597e08a3b9acbd2982c/addon/components/sl-calendar-day.js#L18)):*
Requires a DocBlock

---

*[addon/components/sl-calendar-day.js, line 26 \[r2\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-K4hLHf7EQu3ty0OTfTI:-K4hLHf7EQu3ty0OTfTJ:-880895768) ([raw file](https://github.com/softlayer/sl-ember-components/blob/88bd9102a242a3b5b93cf597e08a3b9acbd2982c/addon/components/sl-calendar-day.js#L26)):*
Remove trailing comma

---

*[addon/components/sl-calendar-day.js, line 37 \[r2\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-K4hLG_sQvIIrrYBF93T:-K4hLG_sQvIIrrYBF93U:101773398) ([raw file](https://github.com/softlayer/sl-ember-components/blob/88bd9102a242a3b5b93cf597e08a3b9acbd2982c/addon/components/sl-calendar-day.js#L37)):*
* Belongs in the "Properties" comment block and requires a DocBlock
* What purpose is this serving?

---

*[addon/templates/components/sl-calendar.hbs, line 48 \[r2\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-K4hLZitWLHH6LnAi7nU:-K4hLZitWLHH6LnAi7nV:-2130312955) ([raw file](https://github.com/softlayer/sl-ember-components/blob/88bd9102a242a3b5b93cf597e08a3b9acbd2982c/addon/templates/components/sl-calendar.hbs#L48)):*
Why the style declaration?  It's not present in the "viewing days" condition.  If needs to be present can we do it via css declaration rather than inline?

---

*[addon/templates/components/sl-calendar.hbs, line 51 \[r2\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-K4hLkugm8sih1T5Z2n0:-K4hLkuh7uzzIAm-c4EQ:1399985408) ([raw file](https://github.com/softlayer/sl-ember-components/blob/88bd9102a242a3b5b93cf597e08a3b9acbd2982c/addon/templates/components/sl-calendar.hbs#L51)):*
Nitpicky: be consistent in the order used for "class" and "collapse" - the "viewing days" condition has a different order for similar DOM structure.  Makes is easier to move through the file, compare, etc.

---

*[addon/templates/components/sl-calendar.hbs, line 54 \[r2\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-K4hLi9Q1IgtutkMHT3c:-K4hLi9Q1IgtutkMHT3d:-2130312955) ([raw file](https://github.com/softlayer/sl-ember-components/blob/88bd9102a242a3b5b93cf597e08a3b9acbd2982c/addon/templates/components/sl-calendar.hbs#L54)):*
Why the style declaration?  It's not present in the "viewing days" condition.  If needs to be present can we do it via css declaration rather than inline?

---

*[addon/templates/components/sl-calendar.hbs, line 82 \[r2\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-K4hLzy7QDIwp5JEgUgB:-K4hLzy7QDIwp5JEgUgC:-2130312955) ([raw file](https://github.com/softlayer/sl-ember-components/blob/88bd9102a242a3b5b93cf597e08a3b9acbd2982c/addon/templates/components/sl-calendar.hbs#L82)):*
Why the style declaration?  It's not present in the "viewing days" condition.  If needs to be present can we do it via css declaration rather than inline?

---

*[addon/templates/components/sl-calendar.hbs, line 88 \[r2\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-K4hM0KNG2B4v6ggN9pl:-K4hM0KNG2B4v6ggN9pm:-2130312955) ([raw file](https://github.com/softlayer/sl-ember-components/blob/88bd9102a242a3b5b93cf597e08a3b9acbd2982c/addon/templates/components/sl-calendar.hbs#L88)):*
Why the style declaration?  It's not present in the "viewing days" condition.  If needs to be present can we do it via css declaration rather than inline?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-:-K4hOAu7xUjicMqmTDi2:-1668839031)*

notmessenger:
Build is failing

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-:-K4hTt_znrg7V7QQcXqu:618820389)*

azizpunjani:
Review status: 3 of 6 files reviewed at latest revision, 7 unresolved discussions.

---

*[addon/templates/components/sl-calendar.hbs, line 48 \[r2\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-K4hLZitWLHH6LnAi7nU:-K4i2UOsMu1I3zoxRdCh:235735062) ([raw file](https://github.com/softlayer/sl-ember-components/blob/88bd9102a242a3b5b93cf597e08a3b9acbd2982c/addon/templates/components/sl-calendar.hbs#L48)):*
Was there from before. Looks like dead code. Removed it.

---

*[addon/templates/components/sl-calendar.hbs, line 54 \[r2\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-K4hLi9Q1IgtutkMHT3c:-K4i2_hbc94PpXU_Ez68:235735062) ([raw file](https://github.com/softlayer/sl-ember-components/blob/88bd9102a242a3b5b93cf597e08a3b9acbd2982c/addon/templates/components/sl-calendar.hbs#L54)):*
Was there from before. Looks like dead code. Removed it.

---

*[addon/templates/components/sl-calendar.hbs, line 82 \[r2\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-K4hLzy7QDIwp5JEgUgB:-K4i2bgvMVcHzEYRpa4l:235735062) ([raw file](https://github.com/softlayer/sl-ember-components/blob/88bd9102a242a3b5b93cf597e08a3b9acbd2982c/addon/templates/components/sl-calendar.hbs#L82)):*
Was there from before. Looks like dead code. Removed it.

---

*[addon/templates/components/sl-calendar.hbs, line 88 \[r2\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-K4hM0KNG2B4v6ggN9pl:-K4i2chwbxZuwa5dy2iY:235735062) ([raw file](https://github.com/softlayer/sl-ember-components/blob/88bd9102a242a3b5b93cf597e08a3b9acbd2982c/addon/templates/components/sl-calendar.hbs#L88)):*
Was there from before. Looks like dead code. Removed it.

---

*[addon/components/sl-calendar-day.js, line 16 \[r2\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-K4hL8Cb3408SOvkw-dP:-K4i2gwtR6KEN3kfanaZ:-1927419124) ([raw file](https://github.com/softlayer/sl-ember-components/blob/88bd9102a242a3b5b93cf597e08a3b9acbd2982c/addon/components/sl-calendar-day.js#L16)):*
Was experiment with making the calendar more accessible, seems like a bigger task, so reverted changes.

---

*[addon/components/sl-calendar-day.js, line 18 \[r2\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-K4hLDpMqcie3T1H73ME:-K4i2n4tKdrkdvPGDLVs:-1927419124) ([raw file](https://github.com/softlayer/sl-ember-components/blob/88bd9102a242a3b5b93cf597e08a3b9acbd2982c/addon/components/sl-calendar-day.js#L18)):*
Was experiment with making the calendar more accessible, seems like a bigger task, so reverted changes.

---

*[addon/components/sl-calendar-day.js, line 37 \[r2\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-K4hLG_sQvIIrrYBF93T:-K4i2tNbGUzQZJK_S0uD:-1927419124) ([raw file](https://github.com/softlayer/sl-ember-components/blob/88bd9102a242a3b5b93cf597e08a3b9acbd2982c/addon/components/sl-calendar-day.js#L37)):*
Was experiment with making the calendar more accessible, seems like a bigger task, so reverted changes.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209)*

notmessenger:
Reviewed 4 of 4 files at r4.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[addon/templates/components/sl-calendar.hbs, line 8 \[r4\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-K4i5xe5IrS7hPiM9WZl:-K4i5xe5IrS7hPiM9WZm:182081827) ([raw file](https://github.com/softlayer/sl-ember-components/blob/19e2fc53af765203c9433651e39155c77035a983/addon/templates/components/sl-calendar.hbs#L8)):*
Why do we need any content here is not going to give it the `sr-only` class?  Same question applies to other occurrences of this pattern as well.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209)*

azizpunjani:
*[addon/templates/components/sl-calendar.hbs, line 8 \[r4\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-K4i5xe5IrS7hPiM9WZl:-K4iaNkjcXHWgwF3funZ:-1303393325) ([raw file](https://github.com/softlayer/sl-ember-components/blob/19e2fc53af765203c9433651e39155c77035a983/addon/templates/components/sl-calendar.hbs#L8)):*
Content is necessary for screen readers to be able to read aloud. All the `.sr-only` class does is hide the element from plain view. We already [do that](https://github.com/softlayer/sl-ember-components/blob/v0.11.0/app/styles/addon.less#L16) via CSS applied to the icon. The `.sr-only` class is not necessary. The text should still be read aloud by the screen reader because it's present, just not in plain sight.

Of course there are other accessibility issues but that's a whole another task.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209)*

notmessenger:
*[addon/templates/components/sl-calendar.hbs, line 8 \[r4\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-K4i5xe5IrS7hPiM9WZl:-K4iceDrPLzDfMygjJet:-532623083) ([raw file](https://github.com/softlayer/sl-ember-components/blob/19e2fc53af765203c9433651e39155c77035a983/addon/templates/components/sl-calendar.hbs#L8)):*
Your link (https://github.com/softlayer/sl-ember-components/blob/v0.11.0/app/styles/addon.less#L16) is for a right margin.  Is that how it's being hidden?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209)*

azizpunjani:
*[addon/templates/components/sl-calendar.hbs, line 8 \[r4\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-K4i5xe5IrS7hPiM9WZl:-K4ihu1idBHBeVZKas5c:-625631389) ([raw file](https://github.com/softlayer/sl-ember-components/blob/19e2fc53af765203c9433651e39155c77035a983/addon/templates/components/sl-calendar.hbs#L8)):*
Yes, the text is pushed far to the right, then the width is set to 1em and overflow is hidden.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209)*

notmessenger:
*[addon/templates/components/sl-calendar.hbs, line 8 \[r4\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-K4i5xe5IrS7hPiM9WZl:-K4iihK2QjZ_aGfF2NV1:671043014) ([raw file](https://github.com/softlayer/sl-ember-components/blob/19e2fc53af765203c9433651e39155c77035a983/addon/templates/components/sl-calendar.hbs#L8)):*
Why wouldn't we use the Bootstrap-provided mechanism of the `sr-only` class rather than rolling our own approach, even if the CSS is the same (don't know whether it is or not)?  We need to leverage Bootstrap as much as possible.  This class is used in a variety of the other components (https://github.com/softlayer/sl-ember-components/search?utf8=✓&q=only)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209)*

SpikedKira:
*[addon/templates/components/sl-calendar.hbs, line 8 \[r4\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-K4i5xe5IrS7hPiM9WZl:-K4ilus99pefVPQRBkdo:934356620) ([raw file](https://github.com/softlayer/sl-ember-components/blob/19e2fc53af765203c9433651e39155c77035a983/addon/templates/components/sl-calendar.hbs#L8)):*
The sr-only class will remove the dom element from flow (position: absolute) and reduce the height and width to 1px, effectively removing the element entirely from the page as much as possible using CSS.  The idea here is that a screen reader would have CSS off b/c why would they need them at all?  However, our approach leaves the dom element in place so that we can show an icon while hiding relative text that the icon is meant to convey.  Our approach works across the scale instead of just for fully able clients or fully disabled clients.  Having a screen reader turned on while also having CSS turned on should also work, for instance those with limited sight or very poor vision but still not blind.  In this case, highlighting the icon should trigger the client to read aloud "next page" or what have you.  I think this is much preferable to just text or nothing (which would also hide our icon).  

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209)*

notmessenger:
*[addon/templates/components/sl-calendar.hbs, line 8 \[r4\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-K4i5xe5IrS7hPiM9WZl:-K4ipurbbxBGCcVyNCDU:1618581630) ([raw file](https://github.com/softlayer/sl-ember-components/blob/19e2fc53af765203c9433651e39155c77035a983/addon/templates/components/sl-calendar.hbs#L8)):*
Do you recommend I create issues to modify all the other uses of =`sl-only` as well?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209)*

SpikedKira:
*[addon/templates/components/sl-calendar.hbs, line 8 \[r4\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-K4i5xe5IrS7hPiM9WZl:-K4ir_fjDXdLq9_hs4Nq:-1265898386) ([raw file](https://github.com/softlayer/sl-ember-components/blob/19e2fc53af765203c9433651e39155c77035a983/addon/templates/components/sl-calendar.hbs#L8)):*
Possibly.  It's going to be case by case really.  sr-only can still be useful if we have some sort of detail text that explains something that would not be present w/o CSS.  Possibly something like a graph?  For icons/symbols, I think it's not the right solution.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209)*

azizpunjani:
*[addon/templates/components/sl-calendar.hbs, line 8 \[r4\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-K4i5xe5IrS7hPiM9WZl:-K4iwQJvQJbY5idzrqIB:1445153799) ([raw file](https://github.com/softlayer/sl-ember-components/blob/19e2fc53af765203c9433651e39155c77035a983/addon/templates/components/sl-calendar.hbs#L8)):*
I'd like to mention that the bootstrap class `sr-only` works just as well as our solution. If the screen reader is turned on while CSS is turned on, the screen reader will read aloud the text even though the element with the class `sr-only` is absolutely positioned.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209)*

SpikedKira:
*[addon/templates/components/sl-calendar.hbs, line 8 \[r4\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-K4i5xe5IrS7hPiM9WZl:-K4jAQ7abCO8u7sKmWrh:-265165650) ([raw file](https://github.com/softlayer/sl-ember-components/blob/19e2fc53af765203c9433651e39155c77035a983/addon/templates/components/sl-calendar.hbs#L8)):*
Correct, the screen reader will always read it.  The issue is that the content will not be there if CSS is enabled.  The sr-only class removes the element from flow for people that have CSS turned on so that they don't see it.  We want them to see something (the symbol) but not the text.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209)*

azizpunjani:
*[addon/templates/components/sl-calendar.hbs, line 8 \[r4\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-K4i5xe5IrS7hPiM9WZl:-K4wU0fH0GJXWntBRTec:585722568) ([raw file](https://github.com/softlayer/sl-ember-components/blob/19e2fc53af765203c9433651e39155c77035a983/addon/templates/components/sl-calendar.hbs#L8)):*
`sr-only` is meant for text e.g

```
 <span class="glyphicon glyphicon-ok form-control-feedback" aria-hidden="true"></span>
  <span id="inputSuccess5Status" class="sr-only">(success)</span>
```

The glyph will still show because it does not have the `.sr-only` class. The text however will not be displayed because of the `sr-only` class. 

What it comes down to is, do we want to go with what bootstrap provides us out of the box, or with our own implementation. BS4 also has this class present, I don't see any mention of it being removed/modified, unless I missed something.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209)*

azizpunjani:
*[addon/templates/components/sl-calendar.hbs, line 8 \[r4\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-K4i5xe5IrS7hPiM9WZl:-K4wW-bxUBjwRAOhCmm0:1267900465) ([raw file](https://github.com/softlayer/sl-ember-components/blob/19e2fc53af765203c9433651e39155c77035a983/addon/templates/components/sl-calendar.hbs#L8)):*
I guess the other difference is that in order to use the `sr-only` class we have to wrap the text in another element. The Bootstrap way is a little more work than what we currently have to do.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209)*

notmessenger:
*[addon/templates/components/sl-calendar.hbs, line 8 \[r4\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209#-K4i5xe5IrS7hPiM9WZl:-K4wfdS0QiSFbU5uzXQ0:848386228) ([raw file](https://github.com/softlayer/sl-ember-components/blob/19e2fc53af765203c9433651e39155c77035a983/addon/templates/components/sl-calendar.hbs#L8)):*
My vote would be that we do it the Bootstrap way, otherwise why are we using it?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209)*

notmessenger:
Reviewed 3 of 4 files at r5.
Review status: 5 of 6 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1209)*

PR #1392, opened 2016-01-11
Evaluate whether assert.async() needs to be employed in test cases currently using assert.expect()

erangeles:
[Review on Reviewable](https://reviewable.io/reviews/softlayer/sl-ember-components/1392)

### DIFF
--- a/tests/integration/components/sl-button-test.js
+++ b/tests/integration/components/sl-button-test.js
@@ -33,11 +33,15 @@ test( 'Default rendered state', function( assert ) {
 test( 'Default action is triggered when element is clicked', function( assert ) {
     assert.expect( 1 );
 
+    const done = assert.async();
+
     this.on( 'externalAction', () => {
         assert.ok(
             true,
             'External action was called'
         );
+
+        done();
     });
 
     this.render( hbs`
--- a/tests/integration/components/sl-calendar-test.js
+++ b/tests/integration/components/sl-calendar-test.js
@@ -331,6 +331,8 @@ test( 'Action fires when day is clicked', function( assert ) {
 
     assert.expect( 1 );
 
+    const done = assert.async();
+
     this.set( 'currentYear', 2022 );
 
     this.set( 'currentMonth', 9 );
@@ -351,6 +353,8 @@ test( 'Action fires when day is clicked', function( assert ) {
             true,
             'The test action was called'
         );
+
+        done();
     });
 
     this.$( '>:first-child' ).find( '.active' ).click();
@@ -359,6 +363,8 @@ test( 'Action fires when day is clicked', function( assert ) {
 skip( 'Action passes through expected objects in content array', function( assert ) {
 
     assert.expect( 8 );
+
+    const done = assert.async();
 
     this.set( 'currentYear', 2022 );
 
@@ -428,6 +434,8 @@ skip( 'Action passes through expected objects in content array', function( asser
                 'The label property was passed through'
             );
         }
+
+        done();
     });
 
     this.$( '>:first-child' ).find( '.active' ).click();
--- a/tests/integration/components/sl-drop-button-test.js
+++ b/tests/integration/components/sl-drop-button-test.js
@@ -182,6 +182,8 @@ test( 'sl-drop-option label is supported', function( assert ) {
 test( 'Click action triggers bound action', function( assert ) {
     assert.expect( 1 );
 
+    const done = assert.async();
+
     this.render( hbs`
         {{#sl-drop-button}}
             {{sl-drop-option action="testAction" label="red"}}
@@ -193,6 +195,8 @@ test( 'Click action triggers bound action', function( assert ) {
             true,
             'The test action was called'
         );
+
+        done();
     });
 
     this.$( '>:first-child' ).find( 'a' ).click();
--- a/tests/integration/components/sl-drop-option-test.js
+++ b/tests/integration/components/sl-drop-option-test.js
@@ -105,6 +105,8 @@ test( 'if label is present and icon is set icon image tag is rendered with sampl
 test( 'Action is wired into template hyperlink tag', function( assert ) {
     assert.expect( 1 );
 
+    const done = assert.async();
+
     this.render( hbs`
         {{sl-drop-option action="testAction" label="test"}}
     ` );
@@ -114,6 +116,8 @@ test( 'Action is wired into template hyperlink tag', function( assert ) {
             true,
             'The test action was called'
         );
+
+        done();
     });
 
     this.$( '>:first-child' ).find( 'a' ).click();
--- a/tests/integration/components/sl-menu-item-test.js
+++ b/tests/integration/components/sl-menu-item-test.js
@@ -104,6 +104,8 @@ test( '"contains-dropdown" and "dropdown-toggle" classes are set when menu has s
 test( 'clickLink() action is fired from top level menu item', function( assert ) {
     assert.expect( 1 );
 
+    const done = assert.async();
+
     this.set( 'menuItem', menuItem );
 
     this.render( hbs`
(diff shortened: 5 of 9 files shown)


azizpunjani:
I'm not sure we need these `assert.async` calls. If you look at the Ember testing guide, you will notice they do *not* use them for testing either, [user interaction](https://guides.emberjs.com/v2.2.0/testing/testing-components/#toc_testing-user-interaction) or [testing actions](https://guides.emberjs.com/v2.2.0/testing/testing-components/#toc_testing-actions).

Not saying there's no use case for `assert.async`, `sl-modal` and `sl-tab-panel` is an example of when we need to use `assert.async` because of the modal's transition. I don't think we should be using it across the board if it's not needed.

azizpunjani:
Reviewed 9 of 9 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1392)*

notmessenger:
@azizpunjani  If you were to wrap the assertion inside of a `.on()` in an `Ember.run.later()` the test case will still pass even though the assertion never ran.  The use of `assert.async` makes sure that the test case waits for the assertion to be ran.  This doesn't manifest itself because the code runs so quickly sequentially that it is not observed.  However, anything that causes this execution to slow down - unlikely in the test logic itself because of what is being written, so anything within the firing of the `on` event - will cause this scenario to be encountered.

Perhaps too defensive?  If so balance it with a consistent pattern for async code - which this is - versus having to understand which async code won't likely cause you a problem versus that which will.

Thoughts?

---

Reviewed 9 of 9 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[tests/unit/components/sl-calendar-year-test.js, line 52 \[r1\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1392#-K7wdBP2cPOiX_dUOMuI:-K7wdBP2cPOiX_dUOMuJ:1550713398) ([raw file](https://github.com/softlayer/sl-ember-components/blob/712a98e0bf3fa62ca1900940baae170286c05a0e/tests/unit/components/sl-calendar-year-test.js#L52)):*
This should be the first entry in the test - https://github.com/softlayer/ember-style-guide/blob/master/ember-testing.md#assertexpect

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1392#-:-K7wckbLFI2cx7TB3pCQ:284262352)*

erangeles:
Review status: 8 of 9 files reviewed at latest revision, 1 unresolved discussion.

---

*[tests/unit/components/sl-calendar-year-test.js, line 52 \[r1\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1392#-K7wdBP2cPOiX_dUOMuI:-K8Hkvo_7f4SMHi9Q86u:-469593337) ([raw file](https://github.com/softlayer/sl-ember-components/blob/712a98e0bf3fa62ca1900940baae170286c05a0e/tests/unit/components/sl-calendar-year-test.js#L52)):*
done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1392)*

notmessenger:
@azizpunjani Thoughts?

---

Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1392#-:-K8I0NgxBVMCulJKgxN_:-1144356645)*

azizpunjani:
@notmessenger keep in mind I'm saying, we don't need the `assert.async` code, we *would* still have the `assert.expect(..)` line. With the `assert.expect(..)` present in the test, if you were to wrap the assert within the `on` callback in an `Ember.run.later` the `assert` would still run at a late time. If the assertion were not to run, the test would fail.

It seems black and white to me, if a test were passing sometimes and failing at times, then it's a timing issue and we should be looked into and the remedy would likely be using `assert.async`. I think since these cases are not common we should not enforce `assert.async` code unless it's necessary.

notmessenger:
@azizpunjani It is not true that you can have `assert.expect(...) present, wrapped code with `Ember.run.later`, and `assert` would still run at a later time.  Below are a variety of scenarios I have tested.  Notice #4 which is what you describe which fails when ran:

```
import Ember from 'ember';
import { moduleForComponent, test } from 'ember-qunit';
import hbs from 'htmlbars-inline-precompile';

moduleForComponent( 'sl-button', 'Integration | Component | sl button', {
    integration: true
});

test( '1) No done(), no timeout delay, with assert.expect()', function( assert ) {
    assert.expect( 1 );

    this.on( 'externalAction', () => {
        assert.ok(
            true,
            'External action was called'
        );
    });

    this.render( hbs`
        {{#sl-button action="externalAction"}}
            Default Text
        {{/sl-button}}
    ` );

    this.$( '>:first-child' ).click();
});

test( '2) No done(), no timeout delay, no assert.expect()', function( assert ) {
    this.on( 'externalAction', () => {
        assert.ok(
            true,
            'External action was called'
        );
    });

    this.render( hbs`
        {{#sl-button action="externalAction"}}
            Default Text
        {{/sl-button}}
    ` );

    this.$( '>:first-child' ).click();
});

test( 'FAILS: 3) No done(), with timeout delay, no assert.expect()', function( assert ) {
    this.on( 'externalAction', () => {
        Ember.run.later( function() {
            assert.ok(
                true,
                'External action was called'
            );
        }, 3500 );
    });

    this.render( hbs`
        {{#sl-button action="externalAction"}}
            Default Text
        {{/sl-button}}
    ` );

    this.$( '>:first-child' ).click();
});

test( 'FAILS: 4) No done(), with timeout delay, with assert.expect()', function( assert ) {
    assert.expect( 1 );

    this.on( 'externalAction', () => {
        Ember.run.later( function() {
            assert.ok(
                true,
                'External action was called'
            );
        }, 3500 );
    });

    this.render( hbs`
        {{#sl-button action="externalAction"}}
            Default Text
        {{/sl-button}}
    ` );

    this.$( '>:first-child' ).click();
});

test( '5) With done(), no timeout delay, with assert.expect()', function( assert ) {
    assert.expect( 1 );

    const done = assert.async();

    this.on( 'externalAction', () => {
        assert.ok(
            true,
            'External action was called'
        );

        done();
    });

    this.render( hbs`
        {{#sl-button action="externalAction"}}
            Default Text
        {{/sl-button}}
    ` );

    this.$( '>:first-child' ).click();
});

test( 'FAILS: 6) With done(), with timeout delay, with assert.expect()', function( assert ) {
    assert.expect( 1 );

    const done = assert.async();

    this.on( 'externalAction', () => {
        Ember.run.later( function() {
            assert.ok(
                true,
                'External action was called'
            );
        }, 3500 );

        done();
    });

    this.render( hbs`
        {{#sl-button action="externalAction"}}
            Default Text
        {{/sl-button}}
    ` );

    this.$( '>:first-child' ).click();
});

test( 'FAILS: 7) With done(), with timeout delay, no assert.expect()', function( assert ) {
    const done = assert.async();

    this.on( 'externalAction', () => {
        Ember.run.later( function() {
            assert.ok(
                true,
                'External action was called'
            );
        }, 3500 );

        done();
    });

    this.render( hbs`
        {{#sl-button action="externalAction"}}
            Default Text
        {{/sl-button}}
    ` );

    this.$( '>:first-child' ).click();
});
```

---

Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1392#-:-K8K7kXK5eZj-MWjcbb5:1503884127)*

azizpunjani:
Had a conversation with @notmessenger. It was decided that it's best to use `assert.async` across the board so that the developer does not have to discern situations that `assert.async` needs to be used.

notmessenger:
Reviewed 2 of 2 files at r3.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1392)*

PR #1469, opened 2016-02-09
Fix sl-alert testing to test valid themes

erangeles:
[Review on Reviewable](https://reviewable.io/reviews/softlayer/sl-ember-components/1469)

### DIFF
--- a/tests/integration/components/sl-alert-test.js
+++ b/tests/integration/components/sl-alert-test.js
@@ -38,13 +38,13 @@ test( 'Default rendered state', function( assert ) {
 test( 'Theme property is supported', function( assert ) {
 
     this.render( hbs`
-        {{#sl-alert theme="testValue"}}
+        {{#sl-alert theme="warning"}}
             Warning alert
         {{/sl-alert}}
     ` );
 
     assert.ok(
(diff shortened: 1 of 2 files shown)


juwara0:
These tests are performing correctly. They were written to verify: https://github.com/softlayer/sl-ember-components/blob/7e6c655b485cdcfbd92014d80654a455a2acd675/addon/components/sl-alert.js#L112

However, the titles of the tests and the messages in the asserts are not clear on what is being done. These could stand to be updated.

SpikedKira:
@juwara0 the line you linked to is a thrown warning.  Nothing in the current tests asserts that a warning is thrown.

I see no problems with this PR to correct the fact that our own test improperly uses our enum, however I would suggest adding an additional assert that the warning is thrown when a value not in the enum is provided.

@notmessenger and I discussed the future of our enums at one point and decided that at some point we need to make our enums extendable without the need to extend the entire component.  That way a consuming application can extend the enum, add a new theme name easily, and then add their CSS to support said theme, without needing to touch anything else in a component or group of components.

With that being the end-goal, this PR seems to move in the desired direction.  It perhaps does not fall in line with our desired order of operations, but it does seem valid.

PR #1476, opened 2016-02-09
Modify integration and unit test for sl-alert for better test coverage

erangeles:
[Review on Reviewable](https://reviewable.io/reviews/softlayer/sl-ember-components/1476)

### DIFF
--- a/tests/integration/components/sl-alert-test.js
+++ b/tests/integration/components/sl-alert-test.js
@@ -38,13 +38,13 @@ test( 'Default rendered state', function( assert ) {
 test( 'Theme property is supported', function( assert ) {
 
     this.render( hbs`
-        {{#sl-alert theme="testValue"}}
+        {{#sl-alert theme="warning"}}
             Warning alert
         {{/sl-alert}}
     ` );
 
     assert.ok(
-        this.$( '>:first-child' ).hasClass( 'alert-testValue' ),
+        this.$( '>:first-child' ).hasClass( 'alert-warning' ),
         'Warning theme class is applied'
     );
 });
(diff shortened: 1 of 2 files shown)


notmessenger:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[tests/integration/components/sl-alert-test.js, line 41 \[r1\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1476#-KA7Ksgw1mBJ96Uc-E9Y:-KA7Ksgw1mBJ96Uc-E9Z:-1900116961) ([raw file](https://github.com/softlayer/sl-ember-components/blob/aa3d3ec544c7e681dd8b59cfc40e28aa0b30173c/tests/integration/components/sl-alert-test.js#L41)):*
The value used here should not be one of the theme(s) (values) supported by the component.  The purpose of this test is to see whether or not the `theme` property is used in a certain way, in this case concatenated with some other text.  By using "warning" two things have now occurred: 1) why just "warning"?  why not all of the other values too?  If the answer is that you're not specifically testing "warning" then it leads to 2) by using a value to test with that this component likely has no support for in its logic this test will ensure that during a future refactor logic isn't introduced to hard code the "warning" value or something similar.  By using the value of "testValue" we get the most test coverage for our effort.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1476)*

notmessenger:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[tests/integration/components/sl-alert-test.js, line 41 \[r1\]](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1476#-KA7Ksgw1mBJ96Uc-E9Y:-KAAOIoeDc_ZWo9hEKLe:1613589064) ([raw file](https://github.com/softlayer/sl-ember-components/blob/aa3d3ec544c7e681dd8b59cfc40e28aa0b30173c/tests/integration/components/sl-alert-test.js#L41)):*
After conversation with @juwara0 my comment can be disregarded.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/softlayer/sl-ember-components/1476)*

